PR #3, opened 2020-10-05
DBZ-2611 Provide separate project version for releases

jpechane:
https://issues.redhat.com/browse/DBZ-2611

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -11,6 +11,7 @@
 
     <modelVersion>4.0.0</modelVersion>
     <artifactId>debezium-connector-vitess</artifactId>
+    <version>1.4.0-SNAPSHOT</version>
     <name>Debezium Connector for Vitess</name>
     <packaging>jar</packaging>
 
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -36,6 +37,53 @@
         <vitess.vtctld.grpc.port>15999</vitess.vtctld.grpc.port>
     </properties>
 
+    <dependencyManagement>
```
keweishang, line 40: The `debezium-parent` module has already `dependencyManagement`, is it a convention to redo the same thing here?

pom.xml
```diff
@@ -36,6 +37,53 @@
...
+        <dependencies>
+            <!-- Debezium artifacts -->
+            <dependency>
+                <groupId>io.debezium</groupId>
+                <artifactId>debezium-connector-vitess</artifactId>
```
keweishang, line 45: What is the reason to add `debezium-connector-vitess` dependency in `dependencyManagement`?

pom.xml
```diff
@@ -36,6 +37,53 @@
...
+                <version>${project.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>io.debezium</groupId>
+                <artifactId>debezium-connector-mysql</artifactId>
```
keweishang, line 50: We don't need `debezium-connector-mysql` dependency.

pom.xml
```diff
@@ -36,6 +37,53 @@
...
+                <version>${project.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>io.debezium</groupId>
+                <artifactId>debezium-core</artifactId>
+                <version>${version.debezium}</version>
+            </dependency>
+            <dependency>
+                <groupId>io.debezium</groupId>
+                <artifactId>debezium-ddl-parser</artifactId>
```
keweishang, line 60: No need for `debezium-ddl-parser`.

jpechane:
@keweishang Thanks for the review, I reduced the unnecessary dependencies.
The need for dependency management is due to release process. Mven release plugin requires the module to be of `-SNAPHSOT` version but if we don't add the plugin management it will mena that Debezium dependencies will be in `-SNAPSHOT` version which is also disallowed. So we change the parent version to for example `1.4.0.Final`, the module version is 1.4.0-SNAPSHOT and the Debezium dependencies are overridden in dependency management to be of the parent one - the final.

keweishang:
@jpechane Thanks for the explanation of the reason of using dependency management. It is for being aligned with maven release plug-in's behavior, I see.

keweishang (approved):

pom.xml
gunnarmorling: @jpechane, good point, I don't think we need those repeated here?

jpechane: @gunnarmorling Explained in https://github.com/debezium/debezium-connector-vitess/pull/3#issuecomment-703583814

gunnarmorling: Ah, I see. Warrants a comment there in the _pom.xml_. I'll add it and merge. Thx!

PR #4, opened 2020-10-05
DBZ-2618 Exclude build-time-only jars

keweishang:


jpechane:
@keweishang Hi, I think it would be better to use the original assembly and change the dependency scopy in dependency management like
```
<dependencyManagement>
            <dependency>
                <groupId>com.google.code.findbugs</groupId>
                <artifactId>jsr305</artifactId>
                <scope>test</scope>
            </dependency>
...
```
I think this better express the role of dependency in rpoejct and is automatically excluded from assembly too.

gunnarmorling:
Thanks, LGTM. @keweishang, just to be sure, those dependencies don't show up in the resulting archive any longer that way, correct?

gunnarmorling:
@keweishang, could you please always add the link to the Jira issue to a PR (I just did it for this one)? Makes it a tad easier to navigate. Thanks!

gunnarmorling:
Ok, can confirm they are not part of the archive any longer. Thanks all!

keweishang:
@gunnarmorling thanks for adding the link to the Jira issue to this PR. I'll follow this convention for the following PRs 👍 
Yes, I also confirm that the excluded jars are not part of the final archive.

PR #5, opened 2020-10-13
DBZ-2635 Add multiple shardgtids in vgtid

keweishang:
https://issues.redhat.com/browse/DBZ-2635

- One connector instance should be able to subscribe to all shards in a keyspace, it is a prerequisite for supporting vitess reshard operation
- Vitess reshard should not produce duplicated data, nor should data be lost
- When subscribing to a VGTID that contains multiple shard GTIDs, it is indeterministic which shard's events we receive first. So to guarantee at-least-once delivery, when restarting the connector, we restart from the previous VGTID and don't skip events. One exception is when all events from the current VGTID is processed and sent to Kafka, in such case, when restarting the connector, we can restart from the current VGTID.
- In addition to the existing unsharded keyspace, add a multi-shards keyspace to Vitess docker container
- Add integration test for multi-shards keyspace.


keweishang:
@gunnarmorling I don't have the permission to assign the PR review to anyone. Would it be possible for me to have this permission? Thanks.

jpechane:
@keweishang Gunnar is out this week and I don't have the required privileges. When he'll come back we will tune the access right s for this repo

jpechane:
@keweishang Thanks a lot, should be out tomorrow in first 1.4 alpha

gunnarmorling:
> @gunnarmorling I don't have the permission to assign the PR review to anyone. Would it be possible for me to have this permission? Thanks.

Doh, this was an oversight on my end. This should work now, you also can commit to this repo and e.g. merge PRs by someone else. Please let me know if there's still issues.

PR #6, opened 2020-11-11
DBZ-2749 Upgrade vitess docker image to use vitess v8.0.0 GA

keweishang:
https://issues.redhat.com/browse/DBZ-2749

### DIFF
--- a/src/test/java/io/debezium/connector/vitess/VitessConnectorIT.java
+++ b/src/test/java/io/debezium/connector/vitess/VitessConnectorIT.java
@@ -133,9 +133,9 @@ public class VitessConnectorIT extends AbstractVitessConnectorTest {
 
         // apply DDL
         TestHelper.execute("ALTER TABLE numeric_table ADD foo INT default 10;");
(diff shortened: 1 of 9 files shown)


src/test/java/io/debezium/connector/vitess/VitessConnectorIT.java
```diff
@@ -134,8 +134,9 @@ public void shouldOffsetIncrementAfterDDL() throws Exception {
         // apply DDL
         TestHelper.execute("ALTER TABLE numeric_table ADD foo INT default 10;");
         // applying DDL for Vitess version v7.0.1 emits 3 gRPC responses: (VGTID, DDL), (BEGIN, VERSION), (VGTID, COMMIT)
```
rgibaiev, line 136: Shall we remove this comment then?

rgibaiev (approved):

src/test/java/io/debezium/connector/vitess/VitessConnectorIT.java
keweishang: I was hesitating. But I think you're right, no need for this comment. Removed.

PR #7, opened 2020-11-16
DBZ-2578 Add support for using Vitess primary key as Kafka message key.

keweishang:
https://issues.redhat.com/browse/DBZ-2578

A Vitess table's primary key will be used as the Kafka message key, or the unique key if the table does not have a primary key.

We will not support multi-column keys because the `flags` provided by VStream doesn't have sufficient information for multi-column keys.

### DIFF
--- a/src/main/java/io/debezium/connector/vitess/connection/ColumnMetaData.java
+++ b/src/main/java/io/debezium/connector/vitess/connection/ColumnMetaData.java
@@ -17,13 +17,13 @@ import io.debezium.relational.Table;
 public class ColumnMetaData {
     private final String columnName;
     private final VitessType vitessType;
-    private final boolean key;
     private final boolean optional;
+    private final KeyMetaData keyMetaData;
 
-    public ColumnMetaData(String columnName, VitessType vitessType, boolean key, boolean optional) {
+    public ColumnMetaData(String columnName, VitessType vitessType, boolean optional, KeyMetaData keyMetaData) {
         this.columnName = columnName;
         this.vitessType = vitessType;
-        this.key = key;
+        this.keyMetaData = keyMetaData;
         this.optional = optional;
     }
 
@@ -35,11 +35,11 @@ public class ColumnMetaData {
         return vitessType;
     }
 
-    public boolean isKey() {
-        return key;
-    }
-
     public boolean isOptional() {
         return optional;
+    }
+
+    public KeyMetaData getKeyMetaData() {
+        return keyMetaData;
     }
 }
--- a/src/main/java/io/debezium/connector/vitess/connection/KeyMetaData.java
+++ b/src/main/java/io/debezium/connector/vitess/connection/KeyMetaData.java
@@ -1,0 +1,25 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.connector.vitess.connection;
+
+/**
+ * If a column is part of the primary key (including multi-column primary key),
+ * or a column is a single-column unique key
+ */
+public enum KeyMetaData {
+    /**
+     * The column is part of the primary key (including multi-column primary key)
+     */
+    IS_KEY,
+    /**
+     * The column is single-column unique key
+     */
+    IS_UNIQUE_KEY,
+    /**
+     * The column is not part of any key, or the column is part of multi-column unique key
+     */
+    NONE
+}
--- a/src/main/java/io/debezium/connector/vitess/connection/VStreamOutputMessageDecoder.java
+++ b/src/main/java/io/debezium/connector/vitess/connection/VStreamOutputMessageDecoder.java
@@ -10,6 +10,7 @@ import static io.debezium.connector.vitess.connection.ReplicationMessage.Column;
 import java.sql.Types;
 import java.time.Instant;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
 
@@ -22,6 +23,7 @@ import io.debezium.connector.vitess.VitessType;
 import io.debezium.connector.vitess.connection.ReplicationMessage.Operation;
 import io.debezium.relational.ColumnEditor;
 import io.debezium.relational.Table;
+import io.debezium.relational.TableEditor;
 import io.debezium.relational.TableId;
 import io.vitess.proto.Query.Field;
 import io.vitess.proto.Query.Row;
@@ -30,6 +32,10 @@ import binlogdata.Binlogdata;
 
 public class VStreamOutputMessageDecoder implements MessageDecoder {
     private static final Logger LOGGER = LoggerFactory.getLogger(VStreamOutputMessageDecoder.class);
+
+    // See all flags: https://dev.mysql.com/doc/dev/mysql-server/8.0.12/group__group__cs__column__definition__flags.html
+    private static final int PRI_KEY_FLAG = 1 << 1;
+    private static final int UNIQUE_KEY_FLAG = 1 << 2;
 
     private Instant commitTimestamp;
     private String transactionId = null;
@@ -321,11 +327,17 @@ public class VStreamOutputMessageDecoder implements MessageDecoder {
                     if (vitessType.getJdbcId() == Types.OTHER) {
                         LOGGER.error("Cannot resolve JDBC type from vstream type {}", columnType);
                     }
-                    // TODO: Vitess community has recently added pk flag
-                    boolean key = false;
+
+                    KeyMetaData keyMetaData = KeyMetaData.NONE;
+                    if ((field.getFlags() & PRI_KEY_FLAG) != 0) {
+                        keyMetaData = KeyMetaData.IS_KEY;
+                    }
+                    else if ((field.getFlags() & UNIQUE_KEY_FLAG) != 0) {
+                        keyMetaData = KeyMetaData.IS_UNIQUE_KEY;
+                    }
                     boolean optional = true;
 
-                    columns.add(new ColumnMetaData(columnName, vitessType, key, optional));
+                    columns.add(new ColumnMetaData(columnName, vitessType, optional, keyMetaData));
                 }
 
                 Table table = resolveTable(schemaName, tableName, columns);
@@ -335,6 +347,8 @@ public class VStreamOutputMessageDecoder implements MessageDecoder {
     }
 
     private Table resolveTable(String schemaName, String tableName, List<ColumnMetaData> columns) {
+        List<String> pkColumnNames = new ArrayList<>();
+        String uniqueKeyColumnName = null;
         List<io.debezium.relational.Column> cols = new ArrayList<>(columns.size());
         for (ColumnMetaData columnMetaData : columns) {
             ColumnEditor editor = io.debezium.relational.Column.editor()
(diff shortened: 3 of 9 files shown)


src/main/java/io/debezium/connector/vitess/connection/ColumnMetaData.java
```diff
@@ -17,14 +17,19 @@
 public class ColumnMetaData {
     private final String columnName;
     private final VitessType vitessType;
-    private final boolean key;
     private final boolean optional;
+    // key and uniqueKey cannot both be true as the same time.
```
rgibaiev, line 21: // key and uniqueKey. Cannot both be true as the same time.

src/main/java/io/debezium/connector/vitess/connection/ColumnMetaData.java
```diff
@@ -17,14 +17,19 @@
...
+    // If a column is part of the primary key, key is true.
+    // If a column is an unique key (NOT multi-column unique key), uniqueKey is true.
```
rgibaiev, line 23: is a unique key

src/main/java/io/debezium/connector/vitess/connection/VStreamOutputMessageDecoder.java
```diff
@@ -335,15 +342,37 @@ private void handleFieldMessage(Binlogdata.VEvent vEvent) {
     }
 
     private Table resolveTable(String schemaName, String tableName, List<ColumnMetaData> columns) {
+        List<String> pkColumnNames = new ArrayList<>();
+        String uniqueColumnName = null;
```
rgibaiev, line 346: Let's maybe rename it to uniqueKeyColumnName ?

rgibaiev (approved):
Looks good. Left few minor comments though

src/main/java/io/debezium/connector/vitess/connection/ColumnMetaData.java
keweishang: fixed

src/main/java/io/debezium/connector/vitess/connection/VStreamOutputMessageDecoder.java
keweishang: fixed

src/main/java/io/debezium/connector/vitess/connection/ColumnMetaData.java
keweishang: fixed

src/main/java/io/debezium/connector/vitess/connection/ColumnMetaData.java
```diff
@@ -17,14 +17,19 @@
 public class ColumnMetaData {
     private final String columnName;
     private final VitessType vitessType;
-    private final boolean key;
     private final boolean optional;
+    // key and uniqueKey. Cannot both be true as the same time.
```
gunnarmorling, line 21: Given it cannot be both at the same time, I'd suggest to use an enum for this:

```
public enum KeyMetaData { IS_KEY, IS_UNIQUE, NONE }
```

src/main/java/io/debezium/connector/vitess/connection/VStreamOutputMessageDecoder.java
```diff
@@ -322,10 +328,11 @@ private void handleFieldMessage(Binlogdata.VEvent vEvent) {
                         LOGGER.error("Cannot resolve JDBC type from vstream type {}", columnType);
                     }
                     // TODO: Vitess community has recently added pk flag
```
gunnarmorling, line 330: Is this TODO comment still relevant?

src/test/java/io/debezium/connector/vitess/VitessConnectorIT.java
```diff
@@ -333,23 +376,28 @@ private void waitForStreamingRunning() throws InterruptedException {
...
+        return assertInsert(statement, expectedSchemaAndValuesByColumn, TestHelper.TEST_UNSHARDED_KEYSPACE, pkField, false);
     }
 
     private SourceRecord assertInsert(
                                       String statement,
                                       List<SchemaAndValueField> expectedSchemaAndValuesByColumn,
-                                      String database) {
+                                      String database,
+                                      String pkField,
+                                      boolean isMultiShardeds) {
```
gunnarmorling, line 389: Should be "isMultiSharded"? Also worth a comment what it is.

src/test/java/io/debezium/connector/vitess/VitessVerifyRecord.java
```diff
@@ -0,0 +1,41 @@
...
+ */
+public class VitessVerifyRecord extends VerifyRecord {
+
+    /**
+     * Verify that the given {@link SourceRecord} is an INSERT record, and that the key exists.
+     *
+     * @param record the source record; may not be null
+     * @param pkField the single field defining the primary key of the struct; may not be null
+     */
+    public static void isValidInsert(SourceRecord record, String pkField) {
```
gunnarmorling, line 26: Interesting, why can't we use the one from the super class that expects the key value?

src/main/java/io/debezium/connector/vitess/connection/VStreamOutputMessageDecoder.java
```diff
@@ -322,10 +328,11 @@ private void handleFieldMessage(Binlogdata.VEvent vEvent) {
...
-                    boolean key = false;
+                    boolean key = (field.getFlags() & PRI_KEY_FLAG) != 0;
```
gunnarmorling, line 331: I'm wondering how that matches with what you say about flag not exposing enough info for multi-column PKs, given that this flag is obtained from the field?

src/main/java/io/debezium/connector/vitess/connection/VStreamOutputMessageDecoder.java
```diff
@@ -335,15 +342,37 @@ private void handleFieldMessage(Binlogdata.VEvent vEvent) {
     }
 
     private Table resolveTable(String schemaName, String tableName, List<ColumnMetaData> columns) {
+        List<String> pkColumnNames = new ArrayList<>();
```
gunnarmorling, line 345: Also here it seems you prepare for multi-column PKs?

gunnarmorling (commented):
Thanks a lot, @keweishang. A few comments inline, mostly to clarify whether it does support multi-column PKs or not. If it does not indeed, mention of this in the docs PR would be good. Thanks!

src/test/java/io/debezium/connector/vitess/VitessConnectorIT.java
keweishang: I renamed the function parameter to `hasMultipleShards` and added comments 👍

src/main/java/io/debezium/connector/vitess/connection/VStreamOutputMessageDecoder.java
keweishang: We do support multi-column PKs, we also support single-column unique key, however, we don't support multi-column unique key. For example:

```
// Kafka message key = id + int_col
DROP TABLE IF EXISTS comp_pk_table;
CREATE TABLE comp_pk_table
(
    id             BIGINT NOT NULL,
    int_col        INT,
    int_col2        INT,
    PRIMARY KEY (id, int_col)
);

// Kafka message key = int_col3
DROP TABLE IF EXISTS no_pk_multi_comp_unique_keys_table;
CREATE TABLE no_pk_multi_comp_unique_keys_table
(
    id             BIGINT NOT NULL,
    int_col        INT,
    int_col2       INT,
    int_col3       INT,
    int_col4       INT,
    int_col5       INT,
    UNIQUE KEY unique_col_1_2 (int_col, int_col2),
    UNIQUE KEY unique_col_3 (int_col3),
    UNIQUE KEY unique_col_4_5 (int_col4, int_col5)
);
```

keweishang: I just added an integration test case for multi-column PK.

src/main/java/io/debezium/connector/vitess/connection/VStreamOutputMessageDecoder.java
keweishang: Actually, multi-column PKs are supported, single-column unique keys are also supported. Only multi-column unique keys are not supported. The following are some examples of the column [flag](https://dev.mysql.com/doc/dev/mysql-server/8.0.12/group__group__cs__column__definition__flags.html) provided by Vitess:

If a table has a single-column PK, the pk column's flag value in binary format is: `0010`.

If a table has multi-column PK, all columns of the PK has the same flag value in binary format: `0010`.

All single-column unique keys have the same flag value in binary format: `0100`.

The first column in a multi-column unique key has the flag value in binary format: `1000`. The second (and 3rd, etc) column in a multi-column unique key has the flag value in binary format: `0000`. For example:
```
DROP TABLE IF EXISTS t;
CREATE TABLE t
(
    id             BIGINT NOT NULL,
    int_col        INT,
    int_col2       INT,
    int_col3       INT,
    int_col4       INT,
    UNIQUE KEY unique_col_1_2 (int_col, int_col2),
    UNIQUE KEY unique_col_3_4 (int_col3, int_col4)
);
```
`int_col` has flag value `1000`, `int_col2` has flag value `0000`, `int_col3` has flag value `1000`, `int_col4` has flag value `0000`. So there's no easy way to create the Kafka message key from multi-column unique keys.

keweishang: I've just refactored this part by using an enum, hopefully, it is a bit more clear now.

keweishang:
Hi @gunnarmorling, I've answered the inline comments and confirmed that we do support multi-column PKs. I've also updated the [Debezium Documentation PR](https://github.com/debezium/debezium/pull/1955) to highlight that only single-column unique keys are supported, and multi-column unique keys are not supported.

src/main/java/io/debezium/connector/vitess/connection/KeyMetaData.java
```diff
@@ -0,0 +1,19 @@
...
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.connector.vitess.connection;
+
+/**
+ * If a column is part of the primary key (including multi-column primary key),
+ * or a column is a single-column unique key
+ */
+public enum KeyMetaData {
+    // The column is part of the primary key (including multi-column primary key)
```
gunnarmorling, line 13: Small nit: Could you use the JavaDoc style:

```
/**
  * The column is part of the primary key (including multi-column primary key)
  */
```

That way, the comment will be properly rendered e.g. when hovering over the enum member name in an IDE.

src/main/java/io/debezium/connector/vitess/connection/VStreamOutputMessageDecoder.java
gunnarmorling: Gotcha, thx.

src/test/java/io/debezium/connector/vitess/VitessConnectorIT.java
```diff
@@ -290,22 +290,101 @@ public void shouldUsePrevVgtidAsOffsetWhenNoVgtidInGrpcResponse() throws Excepti
...
 
-        SourceRecord record = assertRecordInserted(topicNameFromInsertStmt(INSERT_NUMERIC_TYPES_STMT, TestHelper.TEST_SHARDED_KEYSPACE));
-        // verify the offset has multiple shard gtids
-        assertRecordOffset(record, RecordOffset.fromSourceInfo(record), true);
+        stopConnector();
+        assertConnectorNotRunning();
+    }
+
+    @Test
+    public void shouldUseMultiColumnPkAsRecordKey() throws Exception {
```
gunnarmorling, line 309: Could you add the `@FixFor` annotation?

src/test/java/io/debezium/connector/vitess/VitessVerifyRecord.java
gunnarmorling: Ok, makes sense. Thx.

gunnarmorling:
@keweishang, some more minor comments. Other than that, it's good to go for me. Feel free to merge yourself (or @rgibaiev) once those are addressed. Just ping me and I'll merge the docs PR in core, too. Thanks!

src/main/java/io/debezium/connector/vitess/connection/KeyMetaData.java
keweishang: Thanks. I didn't know about that. Fixed

src/test/java/io/debezium/connector/vitess/VitessConnectorIT.java
keweishang: That's a good suggestion! `@FixFor` added to all the test cases in this PR. I'll remember to do it systematically :)

keweishang:
@gunnarmorling thanks for the comments. I've addressed them and merged this PR. Feel free to merge the docs PR: https://github.com/debezium/debezium/pull/1955 Thanks!

PR #8, opened 2020-11-19
DBZ-2579 Add support for Vitess nullable columns

keweishang:
https://issues.redhat.com/browse/DBZ-2579

Vitess 8 added Nullable information in the `FIELD` event [flags](https://dev.mysql.com/doc/dev/mysql-server/8.0.12/group__group__cs__column__definition__flags.html#ga0b89f761611b41b80de1f7d50f9b6020). We can make a field in the SourceRecord optional accordingly.

Instead of adding additional integration tests, existing integration tests already cover this case.


rgibaiev (approved):
Glad to see that every new addition is a simple update over already existing codebase

keweishang:
Yeah. This one is quite easy :)

PR #11, opened 2020-12-08
DBZ-2836 Invalid column name should fail connector with meaningful message

keweishang:
https://issues.redhat.com/browse/DBZ-2836

### DIFF
--- a/src/main/java/io/debezium/connector/vitess/connection/VStreamOutputMessageDecoder.java
+++ b/src/main/java/io/debezium/connector/vitess/connection/VStreamOutputMessageDecoder.java
@@ -322,7 +322,7 @@ public class VStreamOutputMessageDecoder implements MessageDecoder {
                 List<ColumnMetaData> columns = new ArrayList<>(columnCount);
                 for (short i = 0; i < columnCount; ++i) {
                     Field field = fieldEvent.getFields(i);
-                    String columnName = field.getName();
+                    String columnName = validateColumnName(field.getName(), schemaName, tableName);
                     String columnType = field.getType().name();
                     VitessType vitessType = VitessType.resolve(columnType);
                     if (vitessType.getJdbcId() == Types.OTHER) {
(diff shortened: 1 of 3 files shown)


src/main/java/io/debezium/connector/vitess/connection/VStreamOutputMessageDecoder.java
```diff
@@ -388,4 +388,25 @@ else if (uniqueKeyColumnName != null) {
 
         return tableEditor.create();
     }
+
+    private static String validateColumnName(String columnName, String schemaName, String tableName) {
```
rgibaiev, line 392: All this validation logic applied only when we serialize to AVRO format. I guess it is not the case when JSON is used to serialize events...

keweishang: As discussed offline, Vitess VStreamer's schema reloading [bug](https://vitess.slack.com/archives/C0PQY0PTK/p1607500147157900?thread_ts=1606817216.038500&cid=C0PQY0PTK) can cause the column name in the event to have `@` prefix. In this case, I think it’s better to fail than to send the corrupted row event with `@1` `@2` as column names.

gunnarmorling: Can we distinguish those really invalid column names from ones which just aren't Avro-compatible? For the latter, all other connectors support the [sanitize.field.names](https://debezium.io/documentation/reference/connectors/postgresql.html#postgresql-property-sanitize-field-names) option, which will replace any invalid character when using Avro. Ideally, we'd have the same option and behavior for Vitess, too. Then failing hard only on those bug-caused names would be ideal, I think.

keweishang: @gunnarmorling thanks for the suggestion. I'll double-check if Vitess Connector supports the [sanitize.field.names](sanitize.field.names) option. If not supported, I'll create a ticket to add the support for it.

But indeed, for this specific case, it is a Vitess schema reloading bug that causes the normal column names (e.g. `foo`, `bar`) to be anonymized (e.g. `@1`, `@2`).

keweishang: I'll also document this behavior in the Vitess Documentation once this PR is approved.

rgibaiev:
Agreed with Kewei to merge it into master for now. Once this bug is fixed on Vitess side (we hope it will get fixed), we will remove this logic from connector

rgibaiev (approved):

gunnarmorling:
Sounds good to me; is there anything you wanted to add to the docs, @keweishang? And could you log a follow-up issue for using the name sanitzer logic as known from other connectors? Thx!

gunnarmorling (approved):
Good to merge, waiting only to here back in regards to potential docs update.

keweishang:
Thanks. Today or tomorrow, I'll add the docs to warn users not to use column names with the `@` prefix. I'll also add a follow-up issue for testing (and adding if not supported) the [sanitize.field.names](https://github.com/debezium/debezium-connector-vitess/pull/sanitize.field.names) option, as known for other connectors.

gunnarmorling:
Ok, sweet. We plan to cut the 1.4.0.CR1 release on Wednesday, so those timelines nicely align.

keweishang:
Follow-up DBZ issue for the [sanitize.field.names](https://github.com/debezium/debezium-connector-vitess/pull/sanitize.field.names) option is added.

keweishang:
I've added the documentation in this [PR](https://github.com/debezium/debezium/pull/2014).

PR #15, opened 2021-01-06
DBZ-2892 Align deps with Connect runtime and remove unused

jpechane:
https://issues.redhat.com/browse/DBZ-2892
Depends on https://github.com/debezium/debezium/pull/2038

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -20,6 +20,9 @@
         <!-- Dependencies -->
         <version.vitess.grpc>7.0.0</version.vitess.grpc>
         <version.gson>2.7</version.gson>
+        <version.grpc>1.16.0</version.grpc>
+        <version.joda>2.10.1</version.joda>
+        <version.google.protos>1.17.0</version.google.protos>
 
         <!--
           Specify the properties that will be used for setting up the integration tests' Docker container.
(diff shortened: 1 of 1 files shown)


keweishang:
Hi @jpechane, thanks for aligning the deps with Connect runtime.

May I ask a few questions?

- I see Connect runtime already contains netty deps, so I understand why we use `provided` and the same version of netty deps (netty-handler, netty-common, etc)
- However, I don't understand why we need to explicitly control the version of grpc deps (grpc-core, grpc-netty, etc). Connect runtime doesn't have those deps. Same question for proto-google-common-protos and joda-time, why do we need to explicitly control their versions?

Thanks

jpechane:
@keweishang Hi, this is due to an Red Hat product internal rebuild. When vitess connector is built are Red Hat's product in needs to have dependencies to be rebuilt as well. The rebuilt dependencies use a special version suffixed with `-redhat-000001` etc. and thus we need a way how to override the dependencies when needed.
We are going to provide a BOM (https://issues.redhat.com/browse/DBZ-2145) in near future so from that time the version alignment will be done at BOM level and this change will not be necessary again.

PR #18, opened 2021-02-12
DBZ-2908 Remove DECIMAL string sanitisation for Vitess 9 GA

keweishang:
- https://issues.redhat.com/browse/DBZ-3100
- https://issues.redhat.com/browse/DBZ-2908

Vitess 9.0.0 GA has fixed the bug on the server-side. Also, the Vitess Connector starts to support Vitess 9.0.0 GA from now on.

### DIFF
--- a/src/main/java/io/debezium/connector/vitess/connection/ReplicationMessageColumnValueResolver.java
+++ b/src/main/java/io/debezium/connector/vitess/connection/ReplicationMessageColumnValueResolver.java
@@ -6,15 +6,12 @@
 package io.debezium.connector.vitess.connection;
 
 import java.sql.Types;
-import java.util.regex.Pattern;
 
 import io.debezium.connector.vitess.VitessType;
 import io.vitess.proto.Query;
 
 /** Resolve raw column value to Java value */
 public class ReplicationMessageColumnValueResolver {
-
-    private static Pattern WHITESPACE_PATTERN = Pattern.compile("\\s+");
 
     public static Object resolveValue(
                                       VitessType vitessType, ReplicationMessage.ColumnValue value, boolean includeUnknownDatatypes) {
@@ -35,9 +32,6 @@ public class ReplicationMessageColumnValueResolver {
                     return value.asLong();
                 }
             case Types.VARCHAR:
-                if (Query.Type.DECIMAL.name().equals(vitessType.getName())) {
-                    return WHITESPACE_PATTERN.matcher(value.asString()).replaceAll("");
-                }
                 return value.asString();
             case Types.FLOAT:
                 return value.asFloat();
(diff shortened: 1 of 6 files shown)


rgibaiev (approved):
I guess, would be good to update the documentation to state that we support Vitess starting from version 9 :)
Technically we support version 8 as well, but.. with some minor peculiarities

gunnarmorling:
Ok, so support for 8 would be limited then, as the decimal issue would surface there, correct? Still fine for me to solely focus on the latest and greatest at this point. We might want to ensure more stability later on, though. Could you only accordingly update https://github.com/debezium/debezium.github.io/blob/develop/_data/releases/1.5/series.yml, too? We also could put 8 and 9, if you state the limitation in the docs PR.

keweishang:
>Ok, so support for 8 would be limited then, as the decimal issue would surface there, correct?
Correct, the decimal issue would surface in Vitess 8. They've fixed it in Vitess 9 only. It seems that the Vitess community is solely focusing on the newer versions. They have a fast quarterly major release cycle (e.g. 3 months between Vitess 8.0.0 GA and Vitess 9.0.0 GA).

>Could you only accordingly update https://github.com/debezium/debezium.github.io/blob/develop/_data/releases/1.5/series.yml, too?
I've submitted another PR for this: https://github.com/debezium/debezium/pull/2121

>Still fine for me to solely focus on the latest and greatest at this point. We might want to ensure more stability later on, though.
I agree. Right now, we're simply mirroring Vitess's release fast pace of release cycle. Once their release cycle is getting more stable, we want to ensure more stability later on.

>We also could put 8 and 9, if you state the limitation in the docs PR.
I have an idea. I can create another section in the Vitess Documentation dedicated to limitations when using the Vitess Connector against older versions of Vitess. WDYT? @gunnarmorling @rgibaiev

gunnarmorling:
>  I can create another section in the Vitess Documentation dedicated to
limitations when using the Vitess Connector against older versions of
Vitess.

Yes, I think this would work. For the compatibility matrix on the website,
it might be nice to then say one of the following:

* 8*, 9 (and at the bottom, to explain the "*": See the Vitess connector
documentation (link) for limitations when using the connector with earlier
Vitess versions)
* 9* (and at the bottom, to explain the "*": See the Vitess connector
documentation (link) for limitations when using the connector with earlier
Vitess versions)

WDYT?

>

keweishang:
>* 8*, 9 (and at the bottom, to explain the "*": See the Vitess connector documentation (link) for limitations when using the connector with earlier Vitess versions)

This sounds good to me. I'll update the Vitess Documentation this weekend.

gunnarmorling (approved):

gunnarmorling:
Thx!

PR #19, opened 2021-02-16
DBZ-3115 Add integration test for JSON column

keweishang:
https://issues.redhat.com/browse/DBZ-3115

### DIFF
--- a/src/main/java/io/debezium/connector/vitess/VitessValueConverter.java
+++ b/src/main/java/io/debezium/connector/vitess/VitessValueConverter.java
@@ -10,6 +10,7 @@ import java.time.ZoneOffset;
 import org.apache.kafka.connect.data.Field;
 import org.apache.kafka.connect.data.SchemaBuilder;
 
+import io.debezium.data.Json;
 import io.debezium.jdbc.JdbcValueConverters;
 import io.debezium.jdbc.TemporalPrecisionMode;
 import io.debezium.relational.Column;
@@ -32,6 +33,11 @@ public class VitessValueConverter extends JdbcValueConverters {
     // Get Kafka connect schema from Debebzium column.
     @Override
     public SchemaBuilder schemaBuilder(Column column) {
+        String typeName = column.typeName().toUpperCase();
+        if (matches(typeName, "JSON")) {
+            return Json.builder();
+        }
+
         final SchemaBuilder jdbcSchemaBuilder = super.schemaBuilder(column);
         if (jdbcSchemaBuilder == null) {
             return includeUnknownDatatypes ? SchemaBuilder.bytes() : null;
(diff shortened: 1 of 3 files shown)


rgibaiev (approved):

gunnarmorling (commented):
Thanks! So this should use the `io.debezium.data.Json` logical type (see class of that name in core and where it's used). This might not be the case yet?

keweishang:
That's a good point. Unfortunately, Vitess Connector doesn't export any `io.debezium` semantic type (schema name) yet, same for other column types.

I'll try and add `io.debezium.data.Json` semantic type (schema name) to the JSON column type. I'll ping you when it's ready @gunnarmorling

keweishang:
Currently, we have the following AVRO schema (if we use AVRO converter) for the JSON column:
```
{
    "name": "json_col",
    "type": [
        "null",
        "string"
    ],
    "default": null
}
```

I think what we really want is:
```
{
    "name": "json_col",
    "type": [
        "null",
        {
            "type": "string",
            "connect.version": 1,
            "connect.name": "io.debezium.data.Json"
        }
    ],
    "default": null
}
```

gunnarmorling:
> I think what we really want is ...

Yes; the output should be the same as for a JSON column captured with the MySQL connector. These semantic types allow a consumer to interpret this accordingly, e.g. in this case feed it to a JSON parser optionally.

Good point re usage of semantic types in general, I wasn't aware of that. Leaving it to you whether to do this as part of this PR/issue or separately. Would be good to double-check with the MySQL connector which one it affects, e.g. SET, YEAR, ENUM etc.

keweishang:
>These semantic types allow a consumer to interpret this accordingly, e.g. in this case feed it to a JSON parser optionally.

Thanks for the explanation. That answered my question about the use cases of these semantic types.

I've added the semantic type for the `JSON` column type in this PR. Since we're working on adding supports for ENUM, SET, I'll add semantic types for those types in separate PRs.

This PR is ready to be reviewed now.

gunnarmorling:
Applied, thx!

gunnarmorling:
@keweishang, looks like there's an intermittent test failure introduced by this: https://github.com/debezium/debezium-connector-vitess/runs/1911049611. Ordering in the JSON array isn't stable, so sometimes the assertion fails. You'd probably have to access the specific keys and assert their values. Or, we also could introduce JSONAssert: https://github.com/skyscreamer/JSONassert. May be a bit too much for that case alone, but I could see it being useful elsewhere, and as it is a test-only dependency, we can more easily add it.

keweishang:
@gunnarmorling, thanks for the heads-up and investigation. Indeed it seems that the ordering of the JSON keys isn't stable. I'll fix it today with https://github.com/skyscreamer/JSONassert, I think it's worth it to include this test-only dependency.

PR #30, opened 2021-03-16
DBZ-3271 Centralizing column filter handling

gunnarmorling:
https://issues.redhat.com/browse/DBZ-3271

Requires https://github.com/debezium/debezium/pull/2211/


src/main/java/io/debezium/connector/vitess/VitessDatabaseSchema.java
```diff
@@ -64,12 +64,14 @@ private boolean isFilteredOut(TableId id) {
     }
 
     /** Refreshes the schema content with a table constructed externally */
-    public void refresh(Table table) {
+    @Override
```
jpechane, line 67: This is not properly formatted

gunnarmorling: Weird. I could swear I had done it. Force-pushed now.

PR #44, opened 2021-08-30
DBZ-2525 Generifying exclusion of columns from snapshotting

ani-sha:
https://issues.redhat.com/browse/DBZ-2525


keweishang (approved):

gunnarmorling:
Thanks, @ani-sha!

Thanks for reviewing, @keweishang. To give some context: column filters are applied to snapshot selects now, so that excluded columns are not discarded in the connectors only but are not selected in the first place. While we don't have snapshotting for Vitess yet, this change required some adjustments to the framework interfaces.

PR #48, opened 2021-11-16
DBZ-4295: Support stop_on_reshard flag

sonne5:
Closes https://issues.redhat.com/browse/DBZ-4295

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -22,6 +22,7 @@
           Note that the `dockerhost.ip` property is computed from the IP address of DOCKER_HOST, which will
           work on all platforms. We'll set some of these as system properties during integration testing.
         -->
+        <version.vitess>12.0.0</version.vitess>
         <vitess.image>vitess/test-database</vitess.image>
         <!--suppress UnresolvedMavenProperty -->
         <vitess.vtgate.host>${docker.host.address}</vitess.vtgate.host>
@@ -146,6 +147,7 @@
         <dependency>
             <groupId>io.vitess</groupId>
             <artifactId>vitess-grpc-client</artifactId>
+            <version>${version.vitess}</version>
             <exclusions>
                 <exclusion>
                     <groupId>com.google.code.findbugs</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -22,7 +22,6 @@
           Note that the `dockerhost.ip` property is computed from the IP address of DOCKER_HOST, which will
           work on all platforms. We'll set some of these as system properties during integration testing.
         -->
-        <version.vitess>12.0.0</version.vitess>
         <vitess.image>vitess/test-database</vitess.image>
         <!--suppress UnresolvedMavenProperty -->
         <vitess.vtgate.host>${docker.host.address}</vitess.vtgate.host>
@@ -147,7 +146,6 @@
         <dependency>
             <groupId>io.vitess</groupId>
             <artifactId>vitess-grpc-client</artifactId>
-            <version>${version.vitess}</version>
             <exclusions>
                 <exclusion>
                     <groupId>com.google.code.findbugs</groupId>
--- a/src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
+++ b/src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
@@ -82,7 +82,6 @@ public class VitessConnectorConfig extends RelationalDatabaseConnectorConfig {
             .withDisplayName("gtid")
             .withType(Type.STRING)
             .withWidth(Width.LONG)
-            .withDefault("")
             .withImportance(ConfigDef.Importance.HIGH)
             .withDescription(
                     "Single GTID from where to start reading from for a given shard."
@@ -206,7 +205,7 @@ public class VitessConnectorConfig extends RelationalDatabaseConnectorConfig {
         return getConfig().getString(SHARD);
     }
 
-    public String getGTID() {
+    public String getGtid() {
         return getConfig().getString(GTID);
     }
 
--- a/src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
+++ b/src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
@@ -119,6 +119,15 @@ public class VitessConnectorConfig extends RelationalDatabaseConnectorConfig {
                     "Tablet type used to get latest vgtid from Vtctld and get data-changes from Vtgate."
                             + " Value can be MASTER, REPLICA, and RDONLY.");
 
+    public static final Field STOP_ON_RESHARD_FLAG = Field.create(VITESS_CONFIG_GROUP_PREFIX + "stop_on_reshard")
+            .withDisplayName("Tablet type to get data-changes")
+            .withType(Type.BOOLEAN)
+            .withDefault(false)
+            .withWidth(Width.SHORT)
+            .withImportance(ConfigDef.Importance.HIGH)
+            .withDescription("Control StopOnReshard VStream flag."
+                    + " If set true, the old VStream will be stopped after a reshard operation.");
+
     public static final Field INCLUDE_UNKNOWN_DATATYPES = Field.create("include.unknown.datatypes")
             .withDisplayName("Include unknown datatypes")
             .withType(Type.BOOLEAN)
@@ -144,7 +153,8 @@ public class VitessConnectorConfig extends RelationalDatabaseConnectorConfig {
                     VTCTLD_PORT,
                     VTCTLD_USER,
                     VTCTLD_PASSWORD,
-                    TABLET_TYPE)
+                    TABLET_TYPE,
+                    STOP_ON_RESHARD_FLAG)
             .events(INCLUDE_UNKNOWN_DATATYPES)
             .excluding(SCHEMA_EXCLUDE_LIST, SCHEMA_INCLUDE_LIST, PROVIDE_TRANSACTION_METADATA)
             .create();
@@ -220,6 +230,10 @@ public class VitessConnectorConfig extends RelationalDatabaseConnectorConfig {
 
     public String getTabletType() {
         return getConfig().getString(TABLET_TYPE);
+    }
+
+    public boolean getStopOnReshard() {
+        return getConfig().getBoolean(STOP_ON_RESHARD_FLAG);
     }
 
     public boolean includeUnknownDatatypes() {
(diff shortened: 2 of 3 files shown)


Naros:
Hi @sonne5 could you please re-word the commit message to begin with `DBZ-4295`?

sonne5:
> Hi @sonne5 could you please re-word the commit message to begin with `DBZ-4295`?
sure thing, thanks @Naros

keweishang (approved):
LGTM 👍

keweishang:
From the ticket, @sonne5 mentioned:
> If this is `true`, say we are reshard A into B and C, the stream on shard A will be stopped.

If I understood correctly, if this is set to `true`, after resharding, the stream stops on shard A and does not automatically start on shard B and shard A. Essentially, the connector stops streaming from the entire keyspace. How does the connector resume streaming from the given keyspace (e.g. restarting the connector)?

Thanks!

sonne5:
@keweishang thanks for reviewing!

> How does the connector resume streaming from the given keyspace (e.g. restarting the connector)?

So assuming a keyspace with one shard and we have one connector for it. This shard split into 2 shards, with the flag set to true, we need to issue post requests to connect to:
1. stop the connector for the old shard (assuming the old connector won't stop by itself?)
2. start new connectors for the new shards (via different config properties `vitess.shard`)

Does it make sense?

keweishang:
>stop the connector for the old shard (assuming the old connector won't stop by itself?)
Yes. I think the connector won't automatically stop, it simply won't receive new data since the `stop_on_reshard` flag is set to `true`. So you'll need to send a POST request to stop or to remove the old connector.

>start new connectors for the new shards (via different config properties vitess.shard)
You need to be a bit more careful here. You can create two new connectors, one for shard B and one for shard C. Each new connector will start streaming from their corresponding shard. However since they're new connectors, no VGTID offset is stored in the Kafka topic, therefore, they'll start from the latest VGTID for shard B and C. If Shard B and C have any transactions between the moment of the resharding and the moment when the new connectors are created, those transactions won't be streamed to the new connectors. To put it in another way, the new connectors don't have VGTID offset saved and therefore can only receive transactions that happened after the connectors' creation.

sonne5:
> no VGTID offset is stored in the Kafka topic, therefore, they'll start from the latest VGTID for shard B and C. 

This is something I would like get your thoughts as well. I think it is reasonable to ask the reshard workflow to record the cutover GTID somewhere so that in theory we can get to know what's the position that we should start for the new shards after the reshard. 

Looking at the code, the `restartVgtid` is stored inside [SourceInfo](https://github.com/debezium/debezium-connector-vitess/blob/ef0cc917d8348114f60261496f98bac4b721e3cf/src/main/java/io/debezium/connector/vitess/SourceInfo.java#L30) - I'm thinking add another `vitess.gtid` in the config so that if it is not empty we init the vstream request with it (instead of using `CURRENT`: https://github.com/debezium/debezium-connector-vitess/blob/c0d12be3cfe7db29d6cbb1b4c1b258153e681632/src/main/java/io/debezium/connector/vitess/connection/VitessReplicationConnection.java#L216

Does it sound reasonable to you? If so, I can open another PR for that small change.

sonne5:
(^ I cut https://issues.redhat.com/browse/DBZ-4297 for the default starting point)

sonne5:
@keweishang I thought a bit more, it does not make a lot of sense to allow setting `stop_on_resharding = true` while we only allow starting from `current` (latest) position for a shard (the exact reason you called out earlier).

I added another config `vitess.gtid` so that when the shard is set, we allow the starting gtid in the config. If it is not set, we fall back to the vtctld approach.

Let me know if this makes sense to you, happy to discuss more if needed.

keweishang:
>Does it sound reasonable to you? If so, I can open another PR for that small change.

Yes. It sounds reasonable. If I understand correctly, this new configuration `vitess.gtid` will be used only when creating the new connector, after the new connector has been created and running for a while, new transaction's VGITD offset will be saved in Kafka topic and the configuration `vitess.gtid` will not be useful anymore. Sounds good to me. I recommend documenting the new configuration `vitess.gtid` behaviour [here](https://debezium.io/documentation/reference/connectors/vitess.html#vitess-connector-properties), together with the `stop_on_reshard` configuration.

keweishang:
The source code for the Vitess Connector Documentation can be found here: https://github.com/debezium/debezium/blob/main/documentation/modules/ROOT/pages/connectors/vitess.adoc

sonne5:
> this new configuration vitess.gtid will be used only when creating the new connector, after the new connector has been created and running for a while, new transaction's VGITD offset will be saved in Kafka topic and the configuration vitess.gtid will not be useful anymore. 

Yep, you are correct. Happy to update the document after this is merged.

Do you happen to have the permission to submit this PR and do you happen to know what's the release process for this project?

keweishang:
>Do you happen to have the permission to submit this PR and do you happen to know what's the release process for this project?

I haven't been involved in recent releases. @Naros this PR looks good to me. Can you have a final review and merge the PR?

sonne5:
@keweishang unrelated to this PR, just a noob question on vitess connector: how does it handle the restart? Ignore this change, the current behavior is we call vtctld to get current gtid positoin and start the vstream - which seems it is not checking the offset topic.

Intuitively, I would think the thing you described to happen: we only call vtctld to get GTID for the first time, after that we will just consume the offset topic. But looks like we always call VStream endpoint without checking that topic - what do I miss? If you can share some related code pointers, that would be awesome!

pom.xml
```diff
@@ -22,6 +22,7 @@
           Note that the `dockerhost.ip` property is computed from the IP address of DOCKER_HOST, which will
           work on all platforms. We'll set some of these as system properties during integration testing.
         -->
+        <version.vitess>12.0.0</version.vitess>
```
gunnarmorling, line 25: Could you instead please update the version in the [Debezium BOM](https://github.com/debezium/debezium/blob/main/debezium-bom/pom.xml#L26-L31)? All the dependency versions are coordinated via that one.

pom.xml
```diff
@@ -146,6 +147,7 @@
         <dependency>
             <groupId>io.vitess</groupId>
             <artifactId>vitess-grpc-client</artifactId>
+            <version>${version.vitess}</version>
```
gunnarmorling, line 150: Should go away, see above.

src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
```diff
@@ -186,6 +206,10 @@ public String getShard() {
         return getConfig().getString(SHARD);
     }
 
+    public String getGTID() {
```
gunnarmorling, line 209:
```suggestion
    public String getGtid() {
```

Unless it's a static member, we use that style in general for abbreviations in member names (it's not done consistently 100% unfortunately).

src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
```diff
@@ -78,6 +78,16 @@
                             + "E.p. \"0\" for an unsharded keyspace. "
                             + "Or \"-80\" for the -80 shard of the sharded keyspace.");
 
+    public static final Field GTID = Field.create(VITESS_CONFIG_GROUP_PREFIX + "gtid")
+            .withDisplayName("gtid")
+            .withType(Type.STRING)
+            .withWidth(Width.LONG)
+            .withDefault("")
```
gunnarmorling, line 85: I think we can omit the default and just react to the value if it's not `null`?

src/main/java/io/debezium/connector/vitess/connection/VitessReplicationConnection.java
```diff
@@ -223,6 +227,22 @@ public static Vgtid defaultVgtid(VitessConnectorConfig config) {
...
+            final Vgtid gtid = Vgtid.of(
+                    Binlogdata.VGtid.newBuilder()
+                            .addShardGtids(
+                                    Binlogdata.ShardGtid.newBuilder()
+                                            .setKeyspace(config.getKeyspace())
+                                            .setShard(shard)
+                                            .setGtid(startingGtid)
+                                            .build())
+                            .build());
+            LOGGER.info("VGTID '{}' is set to the gtid {} for keyspace: {} shard: {}", gtid, startingGtid, config.getKeyspace(), shard);
```
gunnarmorling, line 243:
```suggestion
            LOGGER.info("VGTID '{}' is set to the GTID {} for keyspace: {} shard: {}", gtid, startingGtid, config.getKeyspace(), shard);
```

gunnarmorling (commented):
Thanks for the change, @sonne5, and thanks for reviewing, @keweishang! Some comments inline. When you do the suggested PR for the BOM change in the main repo, you could also add the docs addition as part of that one. In the context of that main repo PR, this PR's branch will also be built (assuming they are done using the same branch name in both repos).

gunnarmorling:
> What's the release process for this project?

Once those fixes suggested above are done, we can merge the PRs (this repo, and main one), and they will be part of the 1.8.0.Beta1 release, which is scheduled for Nov. 25th.

We're looking for more maintainers for the Vitess connector btw, so if you got more changes in mind, your contributions will be more than welcomed. Thanks a lot!

pom.xml
sonne5: Thanks for the pointer! Since the flag is not available in Vitess 9.0.0 release, we need this PR first https://github.com/debezium/debezium/pull/2951

src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
sonne5: yep, done

pom.xml
gunnarmorling: Merged that PR. Note if both PRs would have used the same branch name, this PR here would have been built against that update over there, without having to merge that other PR first. That's a nice way to avoid merging the upstream PR first and then noticing it caused issues and needs backing out. In this case it was not a problem, as we had seen that the update to 12 works here already. So just mentioning it as some context info :)

sonne5: Ah, didn't know that. TIL

Do you know how do I bump the reference in this branch to pick up the version change? I removed this line and tests are failing.

sonne5: Or is it OK if I remove this two lines in pom.xml and we merge as is?

gunnarmorling: I hope it will pass actually. This build fetches the main branch from the core repo (with your version update) and builds against that (in absence of a branch with the same name over there).

gunnarmorling:
> I do have another idea that we can improve the connector, see some discussion here

Makes sense, PR welcome :)

> looking forward to interacting more on this project!

Same!

Would also love a PR for getting your org added to the [list of Debezium users](https://debezium.io/community/users/), should you be interested 🙏 .

pom.xml
sonne5: Ah yeah, you are right! I guess it's just my local branch issue

gunnarmorling:
Rebased and applied; thx, @sonne5!

PR #50, opened 2021-11-19
DBZ-4324: Remove vtctld dependency in Vitess connector

sonne5:
Resolves https://issues.redhat.com/browse/DBZ-4324


sonne5:
Looks like I don't have the permission to assign reviewers yet, @keweishang @gunnarmorling mind taking a look when you get a chance? Thanks

src/main/java/io/debezium/connector/vitess/connection/VitessConnectionUtils.java
```diff
@@ -0,0 +1,23 @@
...
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.connector.vitess.connection;
+
+import io.debezium.annotation.Immutable;
+
+@Immutable
+public class VitessConnectionUtils {
+    /** The types of vitess tablet. */
+    public enum TabletType {
```
gunnarmorling, line 13: Looks like this could be a top-level type, rather than being nested into the otherwise empty `VitessConnectionUtils`?

gunnarmorling (approved):
One small stylistic comment inline, otherwise this LGTM.

@keweishang, any thoughts from your side, if you have the cycles to spare?

src/main/java/io/debezium/connector/vitess/connection/VitessConnectionUtils.java
keweishang: Yes, looks like this could be a top-level Enum indeed

src/test/java/io/debezium/connector/vitess/VtctldConnection.java
```diff
@@ -26,19 +26,6 @@
 
 /** Use VTCtld to do Vitess admin operations */
 public class VtctldConnection implements AutoCloseable {
```
keweishang, line 28: Is this class `VtctldConnection` still used somewhere? If not, can it be removed as well?

src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
```diff
@@ -82,54 +81,24 @@
             .withDisplayName("gtid")
             .withType(Type.STRING)
             .withWidth(Width.LONG)
+            .withDefault("current")
             .withImportance(ConfigDef.Importance.HIGH)
             .withDescription(
                     "Single GTID from where to start reading from for a given shard."
                             + " It has to be set together with vitess.shard");
 
-    public static final Field VTCTLD_HOST = Field.create(VITESS_CONFIG_GROUP_PREFIX + "vtctld.host")
```
keweishang, line 90: Please remembere to update the configuration documentation as well

keweishang: https://github.com/debezium/debezium/blob/main/documentation/modules/ROOT/pages/connectors/vitess.adoc#vitess-adding-connector-configuration

sonne5:
Addressed comments, ptal @keweishang @gunnarmorling 

Once this is merged, I see https://issues.redhat.com/browse/DBZ-4324 is marked as 1.8.0.Beta1, do we need to worry about this potentially being a breaking change or is it OK since this is a beta release for an incubated project?

gunnarmorling:
Squashed and applied. Thanks a lot, folks!

>  do we need to worry about this potentially being a breaking change or is it OK since this is a beta release for an incubated project?

Nope, we don't have to worry. The Vitess connector is marked as an incubating one, which means we reserve the right to do any breaking changes to config, metrics, message format, etc. On this change in particular, if a user had set those options before, and they upgrade, these now unknown options will simply be ignored. Kafka Connect will just log a warning.

sonne5:
Documentation: https://github.com/debezium/debezium/pull/2961

PR #51, opened 2021-11-24
DBZ-4355: Do not exclude provide.transaction.metadata from config

shichao-an:
Currently provide.transaction.metadata config won't take effect because it is excluded from the config definition.

https://github.com/debezium/debezium-connector-vitess/blob/main/src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java#L134

We need to not exclude it in order to use this this feature.
This fixes https://issues.redhat.com/browse/DBZ-4355

### DIFF
--- a/src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
+++ b/src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
@@ -131,7 +131,7 @@ public class VitessConnectorConfig extends RelationalDatabaseConnectorConfig {
                     TABLET_TYPE,
                     STOP_ON_RESHARD_FLAG)
             .events(INCLUDE_UNKNOWN_DATATYPES)
-            .excluding(SCHEMA_EXCLUDE_LIST, SCHEMA_INCLUDE_LIST, PROVIDE_TRANSACTION_METADATA)
+            .excluding(SCHEMA_EXCLUDE_LIST, SCHEMA_INCLUDE_LIST)
             .create();
 
     /**


sonne5 (approved):
Good catch!

gunnarmorling:
Good spot indeed, wasn't aware that this actually is supported for the Vitess connector already. In fact, we neither seem to have tests nor documentation for this. Would you be up for adding these things as well? There is a test `TransactionMetadataIT` for all the connectors supporting this feature, which can serve as a template for a test in the Vitess connector module.

In terms of docs, there's a section [Transaction metadata](https://debezium.io/documentation/reference/nightly/connectors/postgresql.html#postgresql-transaction-metadata) in the different connector docs pages, and the two options [provide.transaction.metadata](https://debezium.io/documentation/reference/nightly/connectors/postgresql.html#postgresql-property-provide-transaction-metadata) and [transaction.topic](https://debezium.io/documentation/reference/nightly/connectors/postgresql.html#postgresql-property-transaction-topic) should be added to the table with all the options supported by a connector.

jeffchao:
Nice catch. +1 on tests.

shichao-an:
@gunnarmorling Thanks for the comments and pointers on the test suite/docs. Agreed we should add th tests for it.
I am looking at adding `TransactionMetadataIT` for Vitess connector. I think I need to spend some time on getting familiar with the codebase and Vitess test setup (mvn, docker, etc.). 

Since the next release cut is tomorrow, I wonder if we can merge this small change and I can follow up with a separate PR for the tests? We wanted to use transaction meta feature in our recently spun-up pipeline, and this could potentially saves much time.

jeffchao (approved):

gunnarmorling:
> Since the next release cut is tomorrow, I wonder if we can merge this small change and I can follow up with a separate PR for the tests?

Yes, no problem. Could you log a follow-up issue in Jira so we don't forget about this?

In regards to the release, it seems we'll have to move this to early next week unfortunately, as @jpechane is out today.

shichao-an:
@gunnarmorling Sure. I saw you created https://issues.redhat.com/browse/DBZ-4361 . I will use that ticket.

PR #55, opened 2021-12-02
DBZ-4389: Add keepalive pings and handle retriable exceptions

shichao-an:
- Add a new config "vitess.keepalive.interval.ms" which configures the gRPC keepalive interval for VStream.
- Add VitessErrorHandler and marks UNAVAILABLE gRPC status error as retriable.
- Fix a race condition issue for integration tests.

This fixes https://issues.redhat.com/browse/DBZ-4389

### DIFF
--- a/src/main/java/io/debezium/connector/vitess/VitessErrorHandler.java
+++ b/src/main/java/io/debezium/connector/vitess/VitessErrorHandler.java
@@ -1,0 +1,27 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.connector.vitess;
+
+import static io.grpc.Status.Code.UNAVAILABLE;
+
+import io.debezium.connector.base.ChangeEventQueue;
+import io.debezium.pipeline.ErrorHandler;
+import io.grpc.StatusRuntimeException;
+
+public class VitessErrorHandler extends ErrorHandler {
+    public VitessErrorHandler(String logicalName, ChangeEventQueue<?> queue) {
+        super(VitessConnector.class, logicalName, queue);
+    }
+
+    @Override
+    protected boolean isRetriable(Throwable throwable) {
+        if (throwable instanceof StatusRuntimeException) {
+            final StatusRuntimeException exception = (StatusRuntimeException) throwable;
+            return exception.getStatus().getCode().equals(UNAVAILABLE);
+        }
+        return false;
+    }
+}
--- a/src/main/java/io/debezium/connector/vitess/connection/VitessReplicationConnection.java
+++ b/src/main/java/io/debezium/connector/vitess/connection/VitessReplicationConnection.java
@@ -171,6 +171,7 @@ public class VitessReplicationConnection implements ReplicationConnection {
                         .setFlags(vStreamFlags)
                         .build(),
                 responseObserver);
+        LOGGER.info("Started VStream");
     }
 
     private VitessGrpc.VitessStub newStub(ManagedChannel channel) {
@@ -194,6 +195,7 @@ public class VitessReplicationConnection implements ReplicationConnection {
     private ManagedChannel newChannel(String vtgateHost, int vtgatePort) {
         ManagedChannel channel = ManagedChannelBuilder.forAddress(vtgateHost, vtgatePort)
                 .usePlaintext()
+                .keepAliveTime(config.getKeepaliveInterval().toMillis(), TimeUnit.MILLISECONDS)
                 .build();
         return channel;
     }
--- a/src/test/java/io/debezium/connector/vitess/AbstractVitessConnectorTest.java
+++ b/src/test/java/io/debezium/connector/vitess/AbstractVitessConnectorTest.java
@@ -200,8 +200,9 @@ public abstract class AbstractVitessConnectorTest extends AbstractConnectorTest 
         return gtid.substring(0, idx) + seq;
     }
 
-    protected TestConsumer testConsumer(int expectedRecordsCount, String... topicPrefixes) {
-        return new TestConsumer(expectedRecordsCount, topicPrefixes);
+    protected TestConsumer testConsumer(int expectedRecordsCount, String... topicPrefixes) throws InterruptedException {
+        TestConsumer consumer = new TestConsumer(expectedRecordsCount, topicPrefixes);
+        return consumer;
     }
 
     /** Same as io.debezium.connector.postgresql.AbstractRecordsProducerTest.TestConsumer */
--- a/src/test/java/io/debezium/connector/vitess/VitessConnectorIT.java
+++ b/src/test/java/io/debezium/connector/vitess/VitessConnectorIT.java
@@ -559,6 +559,13 @@ public class VitessConnectorIT extends AbstractVitessConnectorTest {
                 .until(() -> logInterceptor.containsMessage("Default VGTID '[{\"keyspace\":"));
     }
 
+    private void waitForVStreamStarted(final LogInterceptor logInterceptor) {
+        // The inserts must happen only after VStream is started with some buffer time.
+        Awaitility.await().atMost(Duration.ofSeconds(TestHelper.waitTimeForRecords()))
+                .pollInterval(Duration.ofSeconds(1))
+                .until(() -> logInterceptor.containsMessage("Started VStream"));
+    }
+
     private void startConnector() throws InterruptedException {
         startConnector(false);
     }
@@ -576,9 +583,11 @@ public class VitessConnectorIT extends AbstractVitessConnectorTest {
     private void startConnector(Function<Configuration.Builder, Configuration.Builder> customConfig, boolean hasMultipleShards)
             throws InterruptedException {
         Configuration.Builder configBuilder = customConfig.apply(TestHelper.defaultConfig(hasMultipleShards));
+        final LogInterceptor logInterceptor = new LogInterceptor();
         start(VitessConnector.class, configBuilder.build());
         assertConnectorIsRunning();
         waitForStreamingRunning();
+        waitForVStreamStarted(logInterceptor);
     }
 
     private void waitForStreamingRunning() throws InterruptedException {
--- a/src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
+++ b/src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
@@ -5,6 +5,9 @@
  */
 package io.debezium.connector.vitess;
 
+import java.time.Duration;
+import java.time.temporal.ChronoUnit;
+
 import org.apache.kafka.common.config.ConfigDef;
 import org.apache.kafka.common.config.ConfigDef.Type;
 import org.apache.kafka.common.config.ConfigDef.Width;
@@ -106,6 +109,15 @@ public class VitessConnectorConfig extends RelationalDatabaseConnectorConfig {
             .withDescription("Control StopOnReshard VStream flag."
                     + " If set true, the old VStream will be stopped after a reshard operation.");
 
+    public static final Field KEEPALIVE_INTERVAL_MS = Field.create(VITESS_CONFIG_GROUP_PREFIX + "keepalive.interval.ms")
+            .withDisplayName("VStream gRPC keepalive interval (ms)")
+            .withType(Type.LONG)
+            .withDefault(Long.MAX_VALUE)
+            .withWidth(Width.SHORT)
+            .withImportance(ConfigDef.Importance.MEDIUM)
+            .withDescription("Control the interval between periodic gPRC keepalive pings for VStream." +
+                    " Defaults to Long.MAX_VALUE (disabled).");
+
     public static final Field INCLUDE_UNKNOWN_DATATYPES = Field.create("include.unknown.datatypes")
             .withDisplayName("Include unknown datatypes")
             .withType(Type.BOOLEAN)
@@ -129,7 +141,8 @@ public class VitessConnectorConfig extends RelationalDatabaseConnectorConfig {
                     VTGATE_USER,
                     VTGATE_PASSWORD,
                     TABLET_TYPE,
-                    STOP_ON_RESHARD_FLAG)
+                    STOP_ON_RESHARD_FLAG,
+                    KEEPALIVE_INTERVAL_MS)
             .events(INCLUDE_UNKNOWN_DATATYPES)
             .excluding(SCHEMA_EXCLUDE_LIST, SCHEMA_INCLUDE_LIST)
             .create();
@@ -197,6 +210,10 @@ public class VitessConnectorConfig extends RelationalDatabaseConnectorConfig {
 
     public boolean getStopOnReshard() {
         return getConfig().getBoolean(STOP_ON_RESHARD_FLAG);
+    }
+
+    public Duration getKeepaliveInterval() {
+        return getConfig().getDuration(KEEPALIVE_INTERVAL_MS, ChronoUnit.MILLIS);
     }
 
     public boolean includeUnknownDatatypes() {
--- a/src/main/java/io/debezium/connector/vitess/VitessConnectorTask.java
+++ b/src/main/java/io/debezium/connector/vitess/VitessConnectorTask.java
@@ -80,7 +80,7 @@ public class VitessConnectorTask extends BaseSourceTask<VitessPartition, VitessO
                     .build();
 
             // saves the exception in the ChangeEventQueue, later task poll() would throw the exception
-            errorHandler = new ErrorHandler(VitessConnector.class, connectorConfig.getLogicalName(), queue);
+            errorHandler = new VitessErrorHandler(connectorConfig.getLogicalName(), queue);
 
             // for metrics
             final VitessEventMetadataProvider metadataProvider = new VitessEventMetadataProvider();
--- a/src/main/java/io/debezium/connector/vitess/VitessStreamingChangeEventSource.java
+++ b/src/main/java/io/debezium/connector/vitess/VitessStreamingChangeEventSource.java
@@ -71,7 +71,7 @@ public class VitessStreamingChangeEventSource implements StreamingChangeEventSou
             }
             if (error.get() != null) {
                 LOGGER.error("Error during streaming", error.get());
-                throw new RuntimeException(error.get());
+                throw error.get();
             }
         }
         catch (Throwable e) {


src/test/java/io/debezium/connector/vitess/AbstractVitessConnectorTest.java
```diff
@@ -200,8 +200,11 @@ protected static String incrementGtid(String gtid, int increment) {
         return gtid.substring(0, idx) + seq;
     }
 
-    protected TestConsumer testConsumer(int expectedRecordsCount, String... topicPrefixes) {
-        return new TestConsumer(expectedRecordsCount, topicPrefixes);
+    protected TestConsumer testConsumer(int expectedRecordsCount, String... topicPrefixes) throws InterruptedException {
+        TestConsumer consumer = new TestConsumer(expectedRecordsCount, topicPrefixes);
+        // Sleep for 1 second before return to avoid race conditions.
+        Thread.sleep(1000);
```
gunnarmorling, line 206: Can we avoid such unconditional thread sleep call? It's not reliable typically and it increases execution time of the test suite. What's the race condition about? Typically, the better fix is to use Awaitility to make sure some specific condition has been matched in a test.

gunnarmorling (commented):
The change itself LGTM. One commen on that sleep call inline though. And the option should be added to the connector docs. Thanks, @shichao-an!

src/main/java/io/debezium/connector/vitess/connection/VitessReplicationConnection.java
```diff
@@ -194,6 +194,7 @@ private int getNumOfRowEvents(Vtgate.VStreamResponse response) {
     private ManagedChannel newChannel(String vtgateHost, int vtgatePort) {
         ManagedChannel channel = ManagedChannelBuilder.forAddress(vtgateHost, vtgatePort)
                 .usePlaintext()
+                .keepAliveTime(config.getKeepaliveInterval().toMillis(), TimeUnit.MILLISECONDS)
```
sonne5, line 197: It is not super clear to me why the channel get closed before this change - for my own understanding, is the channel closed by grpc because it is idle? I think the default idle timeout was 30 min: https://grpc.github.io/grpc-java/javadoc/io/grpc/ManagedChannelBuilder.html#idleTimeout-long-java.util.concurrent.TimeUnit-

src/test/java/io/debezium/connector/vitess/AbstractVitessConnectorTest.java
sonne5: +1

could you also comment a bit around where does the race condition come from? would [stopConnector](https://github.com/debezium/debezium/blob/main/debezium-embedded/src/test/java/io/debezium/embedded/AbstractConnectorTest.java#L146) help to handle that?

sonne5 (commented):
Thanks for working on this @shichao-an 

The change makes sense to me, had a few questions in line. 

Related to this fix: I think we observe the VStream endpoint would be "broken" if there is no events for `net_read_timeout` (the mysql setting). Do we observe the same behavior here? I know you verified that after this change there is no error being raised after some idle time, it might be good to also confirm if we do another write after 10 min, do we still see that event go through.

src/main/java/io/debezium/connector/vitess/VitessErrorHandler.java
```diff
@@ -0,0 +1,28 @@
...
+    }
+
+    @Override
+    protected boolean isRetriable(Throwable throwable) {
+        if (throwable instanceof StatusRuntimeException) {
+            final StatusRuntimeException exception = (StatusRuntimeException) throwable;
+            switch (exception.getStatus().getCode()) {
+                case UNAVAILABLE:
+                    return true;
+            }
```
jeffchao, line 24: nit: Are we expecting to add any new error codes soon? I'm asking because we might be able to just use something like:

```suggestion
            return exception.getStatus().getCode().equals(UNAVAILABLE);
```

src/test/java/io/debezium/connector/vitess/AbstractVitessConnectorTest.java
jeffchao: Others have chimed in, but just wanted to add a thought. The general things you could look out for with sleeps is (1) making unit tests longer than they need to and (2) more importantly, sleeps having potential to make unit tests flaky.

Depending on the code you could do something with Awaitility as Gunnar said like `await().until(<something>)` or some other variant of the API. Another thing is to inject a Clock so we can control/advance them as we need in our tests -- although this could be a larger refactor, for another time (I haven't looked as deeply into this part of the codebase yet).

jeffchao:
@shichao-an  LGTM. Could you also elaborate with a comment (in that same test is fine) on more details of the race condition?

Otherwise, just some minor questions/comments. Thanks!

src/main/java/io/debezium/connector/vitess/connection/VitessReplicationConnection.java
shichao-an: After troubleshooting for quite a bit I couldn't make a conclusion. I tried to set idleTimeout to different values and there was no effect at all. I added debug output and found the channel itself is mostly in IDLE state even before the timeout, so this idle setting might be irrelevant. Because there is only one gRPC activity (VStream) and no new gRPCs. I also debugged server-side as well. 

The general assumption is that the connection/transport is closed after no active streams for a while. To gRPC this means no pings between peers for a period, so the transport is disconnected. I have searched and found these references:
* https://stackoverflow.com/questions/57930529/grpc-connection-use-keepalive-or-idletimeout
* https://stackoverflow.com/questions/64119639/when-should-a-channel-be-closed-on-grpc
* https://grpc.github.io/grpc/core/md_doc_connectivity-semantics-and-api.html
* https://confluence.corp.stripe.com/display/SN/Delayed+Connection+Draining
https://github.com/grpc/grpc/blob/master/doc/keepalive.md
https://cs.mcgill.ca/~mxia3/2019/02/23/Using-gRPC-in-Production/

src/test/java/io/debezium/connector/vitess/AbstractVitessConnectorTest.java
shichao-an: All the Vitess connector's IT tests always failed on my laptop due to similar reason to https://issues.redhat.com/browse/DBZ-4391 . Because the expected data change events could be observed

After troubleshooting for a while, I think there is a race condition. A typical example would be starting from this line https://github.com/debezium/debezium-connector-vitess/blob/main/src/test/java/io/debezium/connector/vitess/VitessConnectorIT.java#L116

When the VStream is started it used "current" as starting gtid position. The test inserts seems started too early when the connector started VStream, so when VTGate receives the connection the insert already completed; this makes VTGate only emit new changes after the insert. This means the insert change event is "lost" for us in the test cases and expecting the record would time out.

shichao-an: If set the GTID to the last GTID in the current Vitess mysql, by passing in configBuilder in the tests, like this:
```
        builder = builder
                .with(RelationalDatabaseConnectorConfig.SERVER_NAME, TEST_SERVER)
                .with(VitessConnectorConfig.VTGATE_HOST, VTGATE_HOST)
                .with(VitessConnectorConfig.VTGATE_PORT, VTGATE_PORT)
                .with(VitessConnectorConfig.VTGATE_USER, USERNAME)
                .with(VitessConnectorConfig.VTGATE_PASSWORD, PASSWORD)
                .with(VitessConnectorConfig.GTID, "MySQL56/0588557d-5357-11ec-84d5-0242ac110002:1-1264")
                .with(VitessConnectorConfig.POLL_INTERVAL_MS, 100);
```
The tests will work. This means it's defintely due to the race condition of VStream with "current"  and inserts as mentioned above.

shichao-an:
> I know you verified that after this change there is no error being raised after some idle time, it might be good to also confirm if we do another write after 10 min, do we still see that event go through.

Yes I tested these cases. If there are writes, the broken time will be delayed, until being idle for the same period of time again when the transport will be disconnected

shichao-an:
I added `waitForVStreamStarted` to wait for VStream started. I still need to add some buffer time using pollInterval. I tried 500ms it worked on my laptop but `shouldSchemaUpdatedAfterOnlineDdl` still failed on GitHub workflow CI. So I changed to to 1s. Let me know if this work for you. @gunnarmorling @sonne5 @jeffchao

src/test/java/io/debezium/connector/vitess/AbstractVitessConnectorTest.java
sonne5: I see, that makes sense. I was using [a slightly different approach](https://github.com/debezium/debezium-connector-vitess/blob/e6aebf00602189d125c88002b5b576b167208732/src/test/java/io/debezium/connector/vitess/VitessReplicationConnectionIT.java#L77-L83) to tackle this exact problem - it was taking advantage that VStream will pass along an event with "OTHER" type when there is no writes happening.

But I think the logging approach you have is more reliable, could you also change it in `VitessReplicationConnectionIT` as well so that we are consistent on that front?

src/test/java/io/debezium/connector/vitess/VitessConnectorIT.java
```diff
@@ -559,6 +559,13 @@ private void waitForShardedGtidAcquiring(final LogInterceptor logInterceptor) {
                 .until(() -> logInterceptor.containsMessage("Default VGTID '[{\"keyspace\":"));
     }
 
+    private void waitForVStreamStarted(final LogInterceptor logInterceptor) {
+        // The inserts must happen only after VStream is started with some buffer time.
+        Awaitility.await().atMost(Duration.ofSeconds(TestHelper.waitTimeForRecords()))
```
sonne5, line 564: nit: add unit suffix here /s/waitTimeForRecords/waitTimeForRecordsSec

src/main/java/io/debezium/connector/vitess/connection/VitessReplicationConnection.java
sonne5: The links seem to suggest using idealTimeout over keepAliveTime 🤷  

I think it's fine to add this option here, the only concern I have here is it requires corresponding setting on the server side as well (e.g., [grpc.keepalive_permit_without_calls](https://grpc.github.io/grpc/core/group__grpc__arg__keys.html#gaf900669f52f137677c4dbb9a7a902c92))

I'm not sure in this case if it worth adding a TODO for it, I will defer to you and others here

sonne5:
> If there are writes, the broken time will be delayed, until being idle for the same period of time again when the transport will be disconnected

just to clairfy, when you say "disconnected" do you mean the `UNAVAILABLE` was being raised or do you mean new writes won't be streamed out? I'm asking because if with this grpc config change, we don't have `UNAVAILABLE` errors, but if the new writes after the certain period of time won't be streamed out - then we have something more to dig into. 

Regardless, this is a good fix anyway 💯 , just want to see if there are more we need to investigate. And since _most_ of the time, we should expect continuous events, the `net_read_timeout` should be something with low priority even if it's real.

src/main/java/io/debezium/connector/vitess/connection/VitessReplicationConnection.java
shichao-an: yeah, idealTimeout doesn't work in my verification so concluded that it's not related to idleTimeout, because idleTimeout is about "channel", but our problem is with "transport/connection", which we need to tackle with keepalive pings.

As for keepalive_permit_without_calls, I tested without this option and it works. According to [this](https://github.com/grpc/grpc/blob/master/doc/keepalive.md) and [this](https://grpc.github.io/grpc-java/javadoc/io/grpc/ManagedChannelBuilder.html#keepAliveWithoutCalls-boolean-):
* allows keepalive pings to be sent even if there are no calls in flight
* Sets whether keepalive will be performed when there are no outstanding RPC on a connection

I think we do have a "call" or outstanding RPC, that is, the VStream call itself. Therefore without this option it won't affect our keepalive pings. I will need to test it in prod for several passes once I had the snapshots ready. We can add it if we do need this option.

shichao-an:
> just to clairfy, when you say "disconnected" do you mean the `UNAVAILABLE` was being raised or do you mean new writes won't be streamed out?
I mean UNAVAILABLE was being raised. After that, new writes won't be streamed out because the connector crashed.

> I'm asking because if with this grpc config change, we don't have `UNAVAILABLE` errors, but if the new writes after the certain period of time won't be streamed out - then we have something more to dig into.

Right, there's no issue with new writes as long as the transport is kept open with keepalive pings. I added the retry behavior to handle UNAVAILABLE error as well, so if somehow the transport is disconnected again when keepalive fails, then connector will be restarted automatically.

sonne5 (approved):
This LGTM, you will also need to update the doc here: https://github.com/debezium/debezium/blob/main/documentation/modules/ROOT/pages/connectors/vitess.adoc

Don't want to block you on the clean up with [VitessReplicationConnectionIT](https://github.com/debezium/debezium-connector-vitess/pull/55#discussion_r761471100), feel free to do that separately

shichao-an:
> This LGTM, you will also need to update the doc here: https://github.com/debezium/debezium/blob/main/documentation/modules/ROOT/pages/connectors/vitess.adoc

Will do in the debezium's PR.

> Don't want to block you on the clean up with [VitessReplicationConnectionIT](https://github.com/debezium/debezium-connector-vitess/pull/55#discussion_r761471100), feel free to do that separately

Sure. We can create a maintenance PR to refactor the IT tests a bit. There is lot of duplicate code here and there.

keweishang:
Looks like I'm late for the party.

I remember that my local docker setup has this issue. Here's the connector error:
```
[2020-06-12 07:40:25,385] INFO No need to process to VGTID (eu.bolt.data.connector.vitess.connection.VStreamOutputMessageDecoder)
[2020-06-12 07:41:14,195] INFO WorkerSourceTask{id=sample_connector-0} Committing offsets (org.apache.kafka.connect.runtime.WorkerSourceTask)
[2020-06-12 07:41:14,195] INFO WorkerSourceTask{id=sample_connector-0} flushing 0 outstanding messages for offset commit (org.apache.kafka.connect.runtime.WorkerSourceTask)
[2020-06-12 07:41:14,202] INFO WorkerSourceTask{id=sample_connector-0} Finished commitOffsets successfully in 7 ms (org.apache.kafka.connect.runtime.WorkerSourceTask)
[2020-06-12 07:42:14,134] INFO WorkerSourceTask{id=sample_connector-0} Committing offsets (org.apache.kafka.connect.runtime.WorkerSourceTask)
[2020-06-12 07:42:14,134] INFO WorkerSourceTask{id=sample_connector-0} flushing 0 outstanding messages for offset commit (org.apache.kafka.connect.runtime.WorkerSourceTask)
[2020-06-12 07:43:14,069] INFO WorkerSourceTask{id=sample_connector-0} Committing offsets (org.apache.kafka.connect.runtime.WorkerSourceTask)
[2020-06-12 07:43:14,069] INFO WorkerSourceTask{id=sample_connector-0} flushing 0 outstanding messages for offset commit (org.apache.kafka.connect.runtime.WorkerSourceTask)
[2020-06-12 07:44:14,002] INFO WorkerSourceTask{id=sample_connector-0} Committing offsets (org.apache.kafka.connect.runtime.WorkerSourceTask)
[2020-06-12 07:44:14,002] INFO WorkerSourceTask{id=sample_connector-0} flushing 0 outstanding messages for offset commit (org.apache.kafka.connect.runtime.WorkerSourceTask)
[2020-06-12 07:45:13,934] INFO WorkerSourceTask{id=sample_connector-0} Committing offsets (org.apache.kafka.connect.runtime.WorkerSourceTask)
[2020-06-12 07:45:13,934] INFO WorkerSourceTask{id=sample_connector-0} flushing 0 outstanding messages for offset commit (org.apache.kafka.connect.runtime.WorkerSourceTask)
[2020-06-12 07:45:53,790] ERROR Producer failure (io.debezium.pipeline.ErrorHandler)
io.grpc.StatusRuntimeException: UNAVAILABLE: Network closed for unknown reason
	at io.grpc.Status.asRuntimeException(Status.java:533)
	at io.grpc.stub.ClientCalls$BlockingResponseStream.hasNext(ClientCalls.java:610)
	at eu.bolt.data.connector.vitess.connection.VitessReplicationConnection$1.read(VitessReplicationConnection.java:67)
	at eu.bolt.data.connector.vitess.VitessStreamingChangeEventSource.execute(VitessStreamingChangeEventSource.java:55)
	at io.debezium.pipeline.ChangeEventSourceCoordinator.lambda$start$0(ChangeEventSourceCoordinator.java:105)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
[2020-06-12 07:45:53,793] INFO Closing replication connection (eu.bolt.data.connector.vitess.connection.VitessReplicationConnection)
[2020-06-12 07:45:53,794] INFO Finished streaming (io.debezium.pipeline.ChangeEventSourceCoordinator)
[2020-06-12 07:45:53,794] INFO Connected metrics set to 'false' (io.debezium.pipeline.metrics.StreamingChangeEventSourceMetrics)
[2020-06-12 07:45:53,983] INFO WorkerSourceTask{id=sample_connector-0} Committing offsets (org.apache.kafka.connect.runtime.WorkerSourceTask)
[2020-06-12 07:45:53,983] INFO WorkerSourceTask{id=sample_connector-0} flushing 0 outstanding messages for offset commit (org.apache.kafka.connect.runtime.WorkerSourceTask)
[2020-06-12 07:45:53,984] ERROR WorkerSourceTask{id=sample_connector-0} Task threw an uncaught and unrecoverable exception (org.apache.kafka.connect.runtime.WorkerTask)
org.apache.kafka.connect.errors.ConnectException: An exception occurred in the change event producer. This connector will be stopped.
	at io.debezium.pipeline.ErrorHandler.setProducerThrowable(ErrorHandler.java:42)
	at eu.bolt.data.connector.vitess.VitessStreamingChangeEventSource.execute(VitessStreamingChangeEventSource.java:88)
	at io.debezium.pipeline.ChangeEventSourceCoordinator.lambda$start$0(ChangeEventSourceCoordinator.java:105)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
Caused by: io.grpc.StatusRuntimeException: UNAVAILABLE: Network closed for unknown reason
	at io.grpc.Status.asRuntimeException(Status.java:533)
	at io.grpc.stub.ClientCalls$BlockingResponseStream.hasNext(ClientCalls.java:610)
	at eu.bolt.data.connector.vitess.connection.VitessReplicationConnection$1.read(VitessReplicationConnection.java:67)
	at eu.bolt.data.connector.vitess.VitessStreamingChangeEventSource.execute(VitessStreamingChangeEventSource.java:55)
	... 6 more
```

And my original Bolt's internal note for this issue was:
```
VStream gRPC server uses the default [keepalive](https://godoc.org/google.golang.org/grpc/keepalive) configuration, which means that the gRPC server doesn’t close the connection and keeps the connection open infinitely. 

We use gRPC client default keepalive configuration too, so it should keep the connection open infinitely too.

The Network closed for unknown reason error happens only when Kafka Connect is within docker-compose. When Kafka Connect is started locally via confluent local start, there is no such problem.

Therefore, it is more of a docker network problem, not a Vitess connector problem.
```

sonne5:
Thanks @keweishang, that's a helpful insights! I think we do run docker compose in production, but in local we are just doing `docker run ... debezium/connect` (shichao can keep me honest here) 

I'm curious if you observed this in local only or in prod as well and how did you solve the problem? Did we just move away from docker compose or did we do something differently?

keweishang:
In local environment, I tried to deploy both Vitess and Connect Worker with the same docker-compose and I had this issue. However, if I ran Connect Worker on bare metal without docker in local environment (I don't remember how did I run Vitess locally in this case, probably in docker or bare metal), I didn't observe this issue.

In production, we've never had this issue, but it's probably due to the fact that there's always events streamed from VTGate and the VStream gRPC is never idle.

keweishang:
Also our production Vitess and Connect Workers are also deployed on bare metal without docker

PR #56, opened 2021-12-06
DBZ-4403: Support blob types and fix uint64 overflows

shichao-an:
Fixes https://issues.redhat.com/browse/DBZ-4403
* Support blob types
* Use string schema (VARCHAR in JDBC) for UINT64 to avoid overflow

### DIFF
--- a/src/main/java/io/debezium/connector/vitess/VitessType.java
+++ b/src/main/java/io/debezium/connector/vitess/VitessType.java
@@ -94,13 +94,14 @@ public class VitessType {
                 return new VitessType(type, Types.BIGINT, resolveEnumAndSetValues(field.getColumnType()));
             case "UINT32":
             case "INT64":
-            case "UINT64":
                 return new VitessType(type, Types.BIGINT);
+            case "UINT64":
             case "VARBINARY":
             case "BINARY":
             case "VARCHAR":
             case "CHAR":
             case "TEXT":
+            case "BLOB":
             case "JSON":
             case "DECIMAL":
             case "TIME":
--- a/src/test/java/io/debezium/connector/vitess/AbstractVitessConnectorTest.java
+++ b/src/test/java/io/debezium/connector/vitess/AbstractVitessConnectorTest.java
@@ -78,8 +78,10 @@ public abstract class AbstractVitessConnectorTest extends AbstractConnectorTest 
             + "text_col,"
             + "mediumtext_col,"
             + "longtext_col,"
+            + "blob_col,"
+            + "mediumblob_col,"
             + "json_col)"
-            + " VALUES ('a', 'bc', '상품 명1', 'リンゴ', 'd', 'ef', 'gh', 'ij', 'kl','mn', '{\"key1\": \"value1\", \"key2\": {\"key21\": \"value21\", \"key22\": \"value22\"}}');";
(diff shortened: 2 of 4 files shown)


src/test/java/io/debezium/connector/vitess/AbstractVitessConnectorTest.java
```diff
@@ -78,8 +78,10 @@
             + "text_col,"
             + "mediumtext_col,"
             + "longtext_col,"
+            + "blob_col,"
+            + "mediumblob_col,"
```
sonne5, line 82: nit: Vitess only supports blob https://github.com/vitessio/vitess/blob/04fd49ee3e5f1b7a6caf01717291954402fd8f0c/go/vt/proto/query/query.pb.go#L270-L272

might worth doing a round of auditing and see if we are missing any other types that are reasonable to sql types: https://github.com/vitessio/vitess/blob/04fd49ee3e5f1b7a6caf01717291954402fd8f0c/go/vt/proto/query/query.pb.go#L210-L309

shichao-an: > nit: Vitess only supports blob.


These two are MySQL types, not Vitess types. Vitess supports all blob types supported by MySQL. 
In Vitess, they (TINYBLOB, BLOB, MEDIUMBLOB and LONGBLOG) are all treated as Blob type ( the `Type_BLOB` in your link). This is similar to TEXT types.

> might worth doing a round of auditing and see if we are missing any other types that are reasonable to sql types:

Besides [those documented](https://debezium.io/documentation/reference/stable/connectors/vitess.html#vitess-basic-types), EXPRESSION and TUPLE ard for internal use. So it should be good in terms of exhaustiveness.

src/main/java/io/debezium/connector/vitess/VitessType.java
```diff
@@ -94,13 +94,14 @@ public static VitessType resolve(Query.Field field) {
                 return new VitessType(type, Types.BIGINT, resolveEnumAndSetValues(field.getColumnType()));
             case "UINT32":
             case "INT64":
-            case "UINT64":
                 return new VitessType(type, Types.BIGINT);
+            case "UINT64":
```
sonne5, line 98: I think we also need to update the [resolver logic](https://github.com/debezium/debezium-connector-vitess/blob/9a87a5ce642d30f2e7f109755ad78916c4ca1227/src/main/java/io/debezium/connector/vitess/connection/ReplicationMessageColumnValueResolver.java#L28-L30) accordingly for this change

sonne5 (commented):
Looks like kafka connect also suggests to [convert uint64 to bytes](https://github.com/a0x8o/kafka/blob/master/connect/api/src/main/java/org/apache/kafka/connect/data/Schema.java#L67-L68) instead of throwing exceptions, /cc @keweishang @gunnarmorling in case there were some other reasons that we choose to throw exception instead of converting them to string originally

The change LGTM I think just need to clean up the resolver a little bit as well.

src/main/java/io/debezium/connector/vitess/VitessType.java
shichao-an: > Looks like kafka connect also suggests to convert uint64 to bytes instead of throwing exceptions
Before this PR, it doesn't throw exceptions upon overflow; it silently become -1, which is a problem.

> I think we also need to update the resolver logic accordingly for this change
Good catch. Will do

shichao-an:
Addressed comments
ptal @sonne5
cc @jeffchao @keweishang

jeffchao:
lgtm.

jeffchao (approved):

sonne5 (approved):

PR #59, opened 2021-12-09
DBZ-4428 Override db and remove schema in Vitess source

sonne5:
Address https://issues.redhat.com/browse/DBZ-4428

## Problem
Follow up on https://github.com/debezium/debezium-connector-vitess/pull/58, we have keyspace information in 3 places in Vitess source: db, keyspace and schema. We should clean it up.

## Solution
This PR remove the `schema` from the Vitess source and override `db` to empty string since in Vitess keyspace is essentially a logical database.

PR to update doc: https://github.com/debezium/debezium/pull/3026

### DIFF
--- a/src/main/java/io/debezium/connector/vitess/VitessSourceInfoStructMaker.java
+++ b/src/main/java/io/debezium/connector/vitess/VitessSourceInfoStructMaker.java
@@ -14,16 +14,13 @@ import io.debezium.connector.AbstractSourceInfoStructMaker;
 public class VitessSourceInfoStructMaker extends AbstractSourceInfoStructMaker<SourceInfo> {
 
     private final Schema schema;
-    private final String keyspace;
 
     public VitessSourceInfoStructMaker(
                                        String connector, String version, VitessConnectorConfig connectorConfig) {
         super(connector, version, connectorConfig);
-        this.keyspace = connectorConfig.getKeyspace();
         this.schema = commonSchemaBuilder()
                 .name("io.debezium.connector.vitess.Source")
                 .field(SourceInfo.KEYSPACE_NAME_KEY, Schema.STRING_SCHEMA)
-                .field(SourceInfo.SCHEMA_NAME_KEY, Schema.STRING_SCHEMA)
                 .field(SourceInfo.TABLE_NAME_KEY, Schema.STRING_SCHEMA)
                 .field(SourceInfo.VGTID_KEY, Schema.STRING_SCHEMA)
                 .build();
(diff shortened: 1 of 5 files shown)


gunnarmorling:
Oh, wow, that was quick. Thanks a lot, @sonne5! So then we actually have this ironed out for CR1 already, very nice 🙏 !

src/test/java/io/debezium/connector/vitess/SourceInfoTest.java
```diff
@@ -96,14 +96,17 @@ public void timestampIsPresent() {
...
+    }
+
+    @Test
+    public void tableIsPresent() {
         assertThat(source.struct().getString(SourceInfo.TABLE_NAME_KEY)).isEqualTo("t");
     }
 
     @Test
-    public void keyspaceIsPresent() {
+    public void keyspaceAndTableArePresent() {
```
shichao-an, line 109: ?

src/main/java/io/debezium/connector/vitess/VitessSourceInfoStructMaker.java
```diff
@@ -37,11 +34,11 @@ public Schema schema() {
     @Override
     public Struct struct(SourceInfo sourceInfo) {
         final Struct res = super.commonStruct(sourceInfo)
-                .put(SourceInfo.KEYSPACE_NAME_KEY, this.keyspace)
-                .put(SourceInfo.SCHEMA_NAME_KEY, sourceInfo.getTableId().schema())
+                .put(SourceInfo.KEYSPACE_NAME_KEY, sourceInfo.database())
```
shichao-an, line 37: I think we'd better update Vitess's `SourceInfo.database()` to empty string and add `SourceInfo.getKeyspace()` and use it here for consistency

jeffchao: FYI @gunnarmorling @sonne5, chiming in here since this is already merged -- just spoke to @shichao-an. This isn't blocking and we'll get it in after the release.

gunnarmorling: No worries. That said, we had an infra issue today, which stopped us from doing the CR1 release as planned, and we'll try again tomorrow (Friday). This is to so, if you wanted to bring in this change, we could try.

If so, feel free to create a PR, but I'd leave it to @jpechane to merge (or not) it before the release, because I am not sure where he is in the process, and whether adding more at this point will throw a wrench between the gears.

sonne5: This is the PR to address comment https://github.com/debezium/debezium-connector-vitess/pull/60

@jeffchao to clarify, I think the PR ^ is just some cosmetic changes, I think the changes that make a differences are:
- https://github.com/debezium/debezium-connector-vitess/pull/58 - which adds `keyspace` in source
- https://github.com/debezium/debezium-connector-vitess/pull/59 - which remove `schema` and `db` in source
they will catch the CR1 release already

shichao-an: Thanks for addressing this. The API change was settled in #59 so we can use it in CR1. The proposed fix #60 is good to have but not required for release.

PR #60, opened 2021-12-09
DBZ-4428 Use keyspace in source info

sonne5:
Address comments in https://github.com/debezium/debezium-connector-vitess/pull/59

- Fix typo in test 
- Add `keyspace()` in SourceInfo and use that directly instead of `database()`

### DIFF
--- a/src/main/java/io/debezium/connector/vitess/SourceInfo.java
+++ b/src/main/java/io/debezium/connector/vitess/SourceInfo.java
@@ -41,6 +41,11 @@ public class SourceInfo extends BaseSourceInfo {
 
     @Override
     protected String database() {
+        // Override DATABASE_NAME_KEY to empty string and in favor of KEYSPACE_NAME_KEY
+        return "";
+    }
+
+    protected String keyspace() {
         return keyspace;
     }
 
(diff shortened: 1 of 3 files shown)


src/main/java/io/debezium/connector/vitess/SourceInfo.java
```diff
@@ -41,6 +41,11 @@ protected Instant timestamp() {
 
     @Override
     protected String database() {
+        // Override DATABASE_NAME_KEY to empty string and in favor of KEYSPACE_NAME_KEY
+        return "";
```
shichao-an, line 45: nit: `return null;`
I saw other connectors do that in some cases and AbstractSourceInfoStructMaker will convert it to "".

shichao-an (approved):
lgtm, minor comment

src/main/java/io/debezium/connector/vitess/SourceInfo.java
sonne5: Sure, done.

it took me a second to remember String can be `null` in Java whereas string is not nil :)

sonne5:
Merging after addressed `"" vs null` comment

PR #61, opened 2021-12-15
DBZ-4458: Add shard attribute to VitessPartition

shichao-an:
Add shard to VitessPartition besides server so the offset commits will be partitioned by (server, shard) when `vitess.shard` is provided in the config.

This fixes https://issues.redhat.com/browse/DBZ-4458

cc @jeffchao @sonne5

### DIFF
--- a/src/main/java/io/debezium/connector/vitess/VitessPartition.java
+++ b/src/main/java/io/debezium/connector/vitess/VitessPartition.java
@@ -14,17 +14,28 @@ import io.debezium.pipeline.spi.Partition;
 import io.debezium.util.Collect;
 
 public class VitessPartition implements Partition {
-    private static final String SERVER_PARTITION_KEY = "server";
+    private static final String SERVER_KEY = "server";
+    private static final String SHARD_KEY = "shard";
 
     private final String serverName;
+    private final String shard;
 
     public VitessPartition(String serverName) {
         this.serverName = serverName;
+        this.shard = null;
+    }
+
+    public VitessPartition(String serverName, String shard) {
+        this.serverName = serverName;
+        this.shard = shard;
     }
 
     @Override
     public Map<String, String> getSourcePartition() {
-        return Collect.hashMapOf(SERVER_PARTITION_KEY, serverName);
+        if (shard == null) {
+            return Collect.hashMapOf(SERVER_KEY, serverName);
+        }
+        return Collect.hashMapOf(SERVER_KEY, serverName, SHARD_KEY, shard);
     }
 
     @Override
@@ -36,12 +47,15 @@ public class VitessPartition implements Partition {
             return false;
         }
         final VitessPartition other = (VitessPartition) obj;
-        return Objects.equals(serverName, other.serverName);
+        return Objects.equals(serverName, other.serverName) && Objects.equals(shard, other.shard);
     }
 
     @Override
     public int hashCode() {
-        return serverName.hashCode();
+        if (shard == null) {
+            return serverName.hashCode();
+        }
+        return Objects.hash(serverName, shard);
     }
 
     static class Provider implements Partition.Provider<VitessPartition> {
@@ -53,7 +67,8 @@ public class VitessPartition implements Partition {
 
         @Override
         public Set<VitessPartition> getPartitions() {
-            return Collections.singleton(new VitessPartition(connectorConfig.getLogicalName()));
+            String shard = connectorConfig.getShard();
+            return Collections.singleton(new VitessPartition(connectorConfig.getLogicalName(), shard));
         }
     }
 }
--- a/src/test/java/io/debezium/connector/vitess/VitessPartitionTest.java
+++ b/src/test/java/io/debezium/connector/vitess/VitessPartitionTest.java
@@ -5,6 +5,10 @@
  */
 package io.debezium.connector.vitess;
 
+import static org.fest.assertions.Assertions.assertThat;
+
+import org.junit.Test;
+
 import io.debezium.connector.common.AbstractPartitionTest;
 
 public class VitessPartitionTest extends AbstractPartitionTest<VitessPartition> {
@@ -17,5 +21,29 @@ public class VitessPartitionTest extends AbstractPartitionTest<VitessPartition> 
     @Override
     protected VitessPartition createPartition2() {
         return new VitessPartition("server2");
+    }
+
+    @Test
+    public void sameShardsShouldBeEqual() {
+        VitessPartition partition1 = new VitessPartition("server1", "shard1");
+        VitessPartition partition2 = new VitessPartition("server1", "shard1");
+        assertThat(partition1).isEqualTo(partition2);
+        assertThat(partition1.hashCode()).isEqualTo(partition2.hashCode());
+    }
+
+    @Test
+    public void differentShardsShouldNotBeEqual() {
+        VitessPartition partition1 = new VitessPartition("server1", "shard1");
+        VitessPartition partition2 = new VitessPartition("server1", "shard2");
+        assertThat(partition1).isNotEqualTo(partition2);
+        assertThat(partition1.hashCode()).isNotEqualTo(partition2.hashCode());
+    }
+
+    @Test
+    public void nullShardShouldNotBeEqualToShard() {
+        VitessPartition partition1 = new VitessPartition("server1");
+        VitessPartition partition2 = new VitessPartition("server1", "shard1");
+        assertThat(partition1).isNotEqualTo(partition2);
+        assertThat(partition1.hashCode()).isNotEqualTo(partition2.hashCode());
     }
 }


jeffchao:
Adding context: Given that `server.name` == "cluster" == same consumer group id here, the problem is each of these connector instances inside the cluster will use a single partition for their `offset.storage.topic` topic. We can get better parallelism if we turn this into a composite key to take advantage of parallelism of that topic.

From Kafka docs:

> offset.storage.topic (default connect-offsets) - topic to use for storing offsets; this topic should have many partitions, be replicated, and be configured for compaction

The setup is a Vitess cluster may have many shards. We have a connector per shard. These connectors sit in the same cluster via consumer group. It would be nice if these connectors could operate independently by reading from different partitions in the cluster's offset topic.

@sonne5 @keweishang @gunnarmorling thoughts?

shichao-an:
I also found that Kafka offset storage actually uses `[connector, partition_map]` as key for the record,
https://github.com/apache/kafka/blob/trunk/connect/runtime/src/main/java/org/apache/kafka/connect/storage/OffsetStorageWriter.java#L147
https://github.com/apache/kafka/blob/trunk/connect/runtime/src/main/java/org/apache/kafka/connect/storage/OffsetStorageReaderImpl.java#L75

If we name our connector by shard, such as `connector:-80`, name it literally achieved what this PR does already.
Therefore I want to collect your ideas about whether we still see value in this PR, because I found similar patterns in mongo and sqlserver, but no mysql and postgres
https://github.com/debezium/debezium/blob/main/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/ReplicaSetPartition.java#L28
https://github.com/debezium/debezium/blob/main/debezium-connector-sqlserver/src/ma[…]in/java/io/debezium/connector/sqlserver/SqlServerPartition.java

jeffchao:
I looked at [DBZ-2975](https://issues.redhat.com/browse/DBZ-2975) and [DBZ-3703](https://issues.redhat.com/browse/DBZ-3703) to get more context.

Using Mongo as an example, it looks like the motivation here is to enable a single connector instance to handle multiple replsets. Within a replset, there are many replicas which a connector will use a tailable cursor with read: majority (in the oplog mode).

This seems a bit different from Vitess, where the equivalent would be a single connector reading from multiple Vitess Keyspaces. Is this right? If so, then this wouldn't be relevant to this use case, but could be useful to be consistent with the multi-partition framework of the other connectors.

shichao-an:
> Using Mongo as an example, it looks like the motivation here is to enable a single connector instance to handle multiple replsets.

Yeah, in sharded cluster mode, the Mongodb will read from multilpe replsets.

> This seems a bit different from Vitess, where the equivalent would be a single connector reading from multiple Vitess Keyspaces.

To me it's more like reading from multiple Vitess shards, because replset is equivalent to shard here.

sonne5:
Looking at how mysql, sqlserver works, it makes me feel Vitess connector is on the opposite side of the problem. A mysql connector is set up to be one connector per node, which can stream out cdc for all the databases in that node; sqlserver can only do that at database level - if we do same mysql setup, it means we need same number of connectors as databases, which _can_ be a lot sometime - hence we have the "partition" concept to partition the data in a node and one connector can handle multiple partitions within the same namespace (i.e., debezium server_name).

Following that logic, I'm not sure the reason that we "partition" mongo events by replsets since within a physical node, there is only one replset already (?)

Going back to Vitess's setup, since there is a middleware between Debezium and sql nodes, ppl usually think keyspace (database) and shards. And the connector has two mode: 1:1 or 1:n (connector vs num_shards). If we follow sqlserver's partition rationale, I think the 1:n mode (where a connector watches multiple shards) is the one that takes "partition" and 1:1 mode is the simple "single partition source" case in Debezium. 

Another thing to note is how Vitess handle 1:n situation in vstream api. The api would return offsets for all the shards as one string and you would pass that string back to Vitess when you want to resume the cdc events. I didn't check sqlserver code, but Vitess is not keeping the offset for all the shards and iterating them one by one. As a result, the 1:n partitioned mode in Vitess also being abstracted as a 1:1 single partitioned source imo. 

I'm +1 to keep Vitess connector consistent with other connectors, but I'm curious how would "partition" concept fit into with server_name in the 1:n use case.

jeffchao:
Yeah, it seems like a different problem, but I'll wait for others with more history 😄 for the motivation.

Specifically for our case, it'll be a trade-off between efficiency and isolation. Given a 1-connector-per-shard for a given keyspace, we would have more connectors but trade-off for better less blast radius via isolation.

Another thing is if we have 1 connector for an entire keyspace serving multiple shards, that might be problematic as we scale to hundreds of shards. The utilization on that single connector could also be choppy as we shard split and merge frequently. That said, users could benefit from a hybrid setup where they have a pool of connectors handling a subset of the overall shards :shipit: .

sonne5:
Yep, that's why we have [stop_on_reshard](https://github.com/debezium/debezium-connector-vitess/pull/48/files) supported in the connector, otherwise Vitess will automatically consume from new shards after shard split even if we initially configured the connector to only watch one shard.

Another interesting thing to consider is given there is a vtgate sit in between and hide the "vitess shard" for us and how it handles the checkpoint offset in 1:N setup, do we still want to expose the "source partition" concept in the Vitess connector. Would love to hear the thoughts from y'all on this as well!

gunnarmorling:
Hey all, thanks for starting this very interesting discussion. Let me first comment on the issue you apparently run into:

> These connectors sit in the same cluster via consumer group.
> ...
> If we name our connector by shard, such as connector:-80, it literally already achieved what this PR does.

So this was confusing me at first, but I think what's happening is that you're using multiple connectors with the same `database.server.name` value. If so, that's definitely not what should be done, as that value must be unique across all the connectors (as stated in [the docs](https://debezium.io/documentation/reference/1.8/connectors/vitess.html#vitess-property-database-server-name)). This is a _logical name_ for identifying different connectors and their artifacts (such as topics written to). While it _might_ make some sense for a sharded source like Vitess, I'd advise against it. If you want to stream multiple shard's change events for one table into a single topic, I'd recommend to use the [logical topic router](https://debezium.io/documentation/reference/stable/transformations/topic-routing.html) for that.

In regards to the partitioning discussion, here's some background on that. Historically, most Debezium connectors only supported a single _source partition_, i.e. a connector would only read from one MySQL or Postgres database, etc. This is reflected by the fact that these connectors only support a single _task_ in terms of Kafka Connect (tasks are means of parallelizing the work a connector does; the source partitions a connector works on are split up between its tasks, which in turn can be shuffled around the worker nodes of a Kafka Connect cluster). This made lots of sense initially, as e.g. reading from the binlog in MySQL is an implicitly serialized operation, and reading the binlog twice, e.g. for two different schemas will only produce overhead.

The exception to that is the connector for MongoDB; as there's implict means of parallelization in MongoDB in the form of replica sets whose oplogs can be tailed independently, that connector supports multiple source partitions (= replica sets) and tasks.

I suppose it may be the same for Vitess due to its notion of keyspaces and shards (I don't quite grok the difference yet). If we can consume multiple change streams from one Vitess "installation" in parallel, then it'd make sense to employ a similar scheme as it's done for MongoDB and its replica sets.

And in fact, as you have seen already via DBZ-2975, there's work happening by @morozov in the context of the SQL Server connector, which will facilitate that. For SQL Server, instead of tailing the TX log, the connector scans _CDC tables_, materialized by SQL Server based on the logs. This operation can be parallelized between multiple logical databases on the same physical SQL Server host. Therefore, @morozov and team are working on making the SQL Server connector (and the Debezium connector framework at large) multi-partition aware: a connector will be able to read from logical databases concurrently, with the load shared between different tasks spawned by the connector. So say there's ten logical databases on one SQL Server host, a connector configured with max.tasks=5 would spin up five tasks, each processing two of the ten databases.

The implementation of this is still in flux; the work has been integrated incrementally into the upstream Debezium repo, but as of now, it's not functional yet (there's many things in place, like ability to configure multiple source databases, management of offsets, etc.). But we're not that far away, and my hope is it can be rolled out as a preview feature in Debezium 1.9 (due end of Q1 '22). There's a pending [design document](https://github.com/debezium/debezium-design-documents/pull/1) which describes the overall approach of this work.

So we should make use of this approach for Vitess eventually too. It'll be a nice validation of the design and framework changes done for the SQL Server connector. But we'd still have to wait a bit until the DBZ-2975 work has been progressed to a state where it's completely functional. It'd help to discuss in the meantime though what the right level of parallelsm for the Vitess connector would be. Does it make sense to read the change streams from multiple keyspaces in parallel? Or rather shards? That's where things are still a bit blurry to me.

shichao-an:
> So this was confusing me at first, but I think what's happening is that you're using multiple connectors with the same database.server.name value. If so, that's definitely not what should be done, as that value must be unique across all the connectors (as stated in the docs). This is a logical name for identifying different connectors and their artifacts (such as topics written to). While it might make some sense for a sharded source like Vitess, I'd advise against it. If you want to stream multiple shard's change events for one table into a single topic, I'd recommend to use the logical topic router for that.

Thanks for clarifying that. I think we haven't invested time in topic routing yet in our prototype yet so we use the same database.server.name as topic prefix to achieve our topic naming. But we do need to use topic routing for that in the future.

> The exception to that is the connector for MongoDB; as there's implict means of parallelization in MongoDB in the form of replica sets whose oplogs can be tailed independently, that connector supports multiple source partitions (= replica sets) and tasks.

Yeah I think it applies to any sharded database.

> I suppose it may be the same for Vitess due to its notion of keyspaces and shards (I don't quite grok the difference yet). If we can consume multiple change streams from one Vitess "installation" in parallel, then it'd make sense to employ a similar scheme as it's done for MongoDB and its replica sets.

Yes. Keyspace is analogous to MongoDB database and shard to replsets. Shard is a subset of data in a keyspace for horizontal partitioning. See https://vitess.io/docs/12.0/reference/features/sharding/

> So we should make use of this approach for Vitess eventually too. It'll be a nice validation of the design and framework changes done for the SQL Server connector. But we'd still have to wait a bit until the DBZ-2975 work has been progressed to a state where it's completely functional. It'd help to discuss in the meantime though what the right level of parallelsm for the Vitess connector would be. Does it make sense to read the change streams from multiple keyspaces in parallel? Or rather shards? That's where things are still a bit blurry to me.

I think this PR doesn't conflict with any of the design and framework changes, and it actually aligns with MongoDB and SQL server's approach ([link1](https://github.com/debezium/debezium/blob/main/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/ReplicaSetPartition.java), [link2](https://github.com/debezium/debezium/blob/main/debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerPartition.java)). This PR can actually benefit @morozov 's recent change on metrics https://github.com/debezium/debezium/pull/2946 and https://github.com/debezium/debezium-connector-vitess/pull/64 by making VitessPartition shard-aware. We wanted to expose shard-level and keyspace-level metrics as a follow-up.

shichao-an:
Hi @gunnarmorling 

cc @jeffchao @y5w 

Just want to bump on this PR and recap a little to previous comments:

* Vitess connector currently supports connector for all shards in a keyspace, OR, one connector per shard in a keyspace. (Because in Vitess the keyspace is the physical isolation of a "database cluster") . But for us, we are interested in one-connector-per-shard model.
* For the new multi-task multi-partition effort, I think it definitely makes sense to leverage multi-task to handle logical databases or shards parallelism. It has more flexibility over the current single-task approach. We will also consider it once it's functional in the future. For Vitess connector, the parallelism should happen at shard-level. As mentioned, the Vitess connector supports either one shard or all shards per connector. So to align with this effort, we can make one Vitess connector to serve multiple shards (instead of 1 or all now). This is similar to MongoDB replsets.
* In terms of this PR which adds the shard in the VitessPartition, I think doesn't conflict with either approach. It has no side effect on offset management either. As mentioned it's also useful for the @morozov 's efforts on metrics which requires Partition classes.

Shall we revisit this PR and wrap up here?

HenryCaiHaiying:
Hi @shichao-an @jeffchao I proposed an approach for multi-task support for Vitess connector in https://issues.redhat.com/browse/DBZ-5382, which might be related to your PR.

HenryCaiHaiying:
If the folks on this thread has no objections on the proposal I have in DBZ-5382, I am going to create a PR to implement that proposal to enable multi-tasks in vitess connector.

jpechane:
@HenryCaiHaiying Hi, is this PR something that is still valid or can be closed?

HenryCaiHaiying:
I feel it can be closed.  But I think @shichao-an is the original author.

jpechane:
Closing for now. @shichao-an Please feel free to re-open if there is still a need for it and we can discuss how to incorporate into the new architecture.

PR #62, opened 2021-12-17
DBZ-4464 Add missing snapshot repo

jpechane:
https://issues.redhat.com/browse/DBZ-4464

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -32,6 +32,22 @@
         <vitess.vtctld.grpc.port>15999</vitess.vtctld.grpc.port>
     </properties>
 
+    <repositories>
+        <repository>
+            <id>oss</id>
+            <name>OSS Sonatype Nexus</name>
+            <url>https://oss.sonatype.org/content/repositories/snapshots</url>
+            <releases>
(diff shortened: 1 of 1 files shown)


gunnarmorling (commented):
Interesting, why does it not suffice to get this from the parent POM?

jpechane:
Because there is no relative path to parent pom and as it is repo separate to the one where parent pom is present you'd presicribe the naming and location of directories.

gunnarmorling:
Ok, so the intention is to suppose a local build of the Vitess repo without having built _debezium/debezium_ before, right? I don't think this should ever happen on CI though, because there each job should build the main repo itself. If that's not the case somewhere, that should be fixed.

gunnarmorling:
In particular, the build failure [here](https://github.com/debezium/debezium/runs/4559043190?check_suite_focus=true) should be handled by having that failing job build the _debezium-embedded module to as part of its build step for the main repo.

gunnarmorling:
Pushed a change to the Vitess workflow in the main repo and this made the failure linked above go away. Hence happy to resolve this one now.

PR #63, opened 2021-12-19
DBZ-4351 Add create/update/delete event seen metrics for monitor upstream dml operation for vitess connector

harveyyue:
Refer https://issues.redhat.com/browse/DBZ-4351


gunnarmorling:
Thx, @harveyyue, had missed that one at first.

PR #66, opened 2022-01-10
DBZ-4532 Add support for extra gRPC headers

shichao-an:
Add support for adding extra gRPC headers in the connector config for VStream.
The "vitess.grpc.headers" config takes value in the form of "k1:v1,k2:v2".

This fixes: https://issues.redhat.com/browse/DBZ-4532

### DIFF
--- a/src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
+++ b/src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
@@ -7,10 +7,15 @@ package io.debezium.connector.vitess;
 
 import java.time.Duration;
 import java.time.temporal.ChronoUnit;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
 
 import org.apache.kafka.common.config.ConfigDef;
 import org.apache.kafka.common.config.ConfigDef.Type;
 import org.apache.kafka.common.config.ConfigDef.Width;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import io.debezium.config.ConfigDefinition;
 import io.debezium.config.Configuration;
@@ -26,6 +31,8 @@ import io.debezium.relational.RelationalDatabaseConnectorConfig;
  * configurations from Debezium.
  */
 public class VitessConnectorConfig extends RelationalDatabaseConnectorConfig {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(VitessConnectorConfig.class);
 
     private static final String VITESS_CONFIG_GROUP_PREFIX = "vitess.";
     private static final int DEFAULT_VTGATE_PORT = 15_991;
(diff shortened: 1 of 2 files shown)


src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
```diff
@@ -216,6 +232,28 @@ public Duration getKeepaliveInterval() {
         return getConfig().getDuration(KEEPALIVE_INTERVAL_MS, ChronoUnit.MILLIS);
     }
 
+    public Map<String, String> getGrpcHeaders() {
```
y5w, line 235: This seems less ideal since we would construct a new map every time we call this. An alternative could be having a private `Map<String, String> grpcHeader` in VitessConnectorConfig and populate like what we do here in the constructor. Then the `getGrpcHeaders` will return that private map directly.

Given this is only being called in one place and we know it will be called once only - I don't feel strong either way, would leave it to you

y5w (commented):
LGTM, I will let Jeff or Gunnar to sign off.

Is there any existing test case that covers this change or did we test that change locally?

jeffchao (approved):
LGTM.

Just a minor question.

src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
shichao-an: Yes. It's called only once during VStream start. I saw couple of places in other *Config class do this (construct a new map in the method) so I follow the same pattern.

src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
```diff
@@ -216,6 +232,28 @@ public Duration getKeepaliveInterval() {
...
+
+        Map<String, String> grpcHeadersMap = new HashMap<>();
+
+        for (String header : grpcHeaders.split(",")) {
+            String[] keyAndValue = header.split(":");
+            if (keyAndValue.length == 2) {
+                grpcHeadersMap.put(keyAndValue[0], keyAndValue[1]);
+            }
+            else {
+                LOGGER.warn("The following gRPC header is invalid: {}", header);
```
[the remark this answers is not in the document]
shichao-an: I intend towards skipping since these are optional headers. I saw a similar pattern for Postgres: https://github.com/debezium/debezium/blob/main/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/PostgresReplicationConnection.java#L721

I will update the Debezium doc for this new config.

shichao-an:
> LGTM, I will let Jeff or Gunnar to sign off.
> 
> Is there any existing test case that covers this change or did we test that change locally?

I tested it locally.

There is ConfigDefinitionMetadataTest extended by most connectors but no concrete tests.

src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
gunnarmorling: Yeah, shouldn't matter much in this case. In general, we tend to extract and build up that config state once in the constructor in cases where this is a more involved operation and simply pass-through to the underlying config if it's a simple read. In that light, I'd probably have extracted that map creation and stored into a field of this class, but as you say it only is called once, so there's no tangible impact at this point.

PR #67, opened 2022-01-12
DBZ-4543 Add stream timeout exception to error handler

shichao-an:
The VStream gRPC may throw a StatusRuntimeException with code=UNKNOWN, description=stream timeout, cause=null.
This error is caused by the stream timeout, when the stream is idle for certain amount of time. Therefore the connector should safely retry upon this error.

This fixes: https://issues.redhat.com/browse/DBZ-4543


jeffchao (approved):
LGTM. Thanks.

src/main/java/io/debezium/connector/vitess/VitessErrorHandler.java
```diff
@@ -20,7 +18,15 @@ public VitessErrorHandler(String logicalName, ChangeEventQueue<?> queue) {
     protected boolean isRetriable(Throwable throwable) {
         if (throwable instanceof StatusRuntimeException) {
             final StatusRuntimeException exception = (StatusRuntimeException) throwable;
-            return exception.getStatus().getCode().equals(UNAVAILABLE);
+            String description = exception.getStatus().getDescription();
```
y5w, line 21: nit: move `description` under the `UNKNOWN` case

src/main/java/io/debezium/connector/vitess/VitessErrorHandler.java
```diff
@@ -20,7 +18,15 @@ public VitessErrorHandler(String logicalName, ChangeEventQueue<?> queue) {
...
+            switch (exception.getStatus().getCode()) {
+                case UNAVAILABLE:
+                    return true;
+                case UNKNOWN:
+                    if (description != null && description.equals("stream timeout")) {
+                        return true;
+                    }
```
y5w, line 28: nit: Might worth adding a comment saying why this should be retryable (basically what you have in the PR description :))

y5w (approved):
🚢

PR #68, opened 2022-01-13
DBZ-2224 Test logging based on logback

jpechane:
Depends on https://github.com/debezium/debezium/pull/3103


shichao-an (approved):
lgtm

gunnarmorling:
Hey @shichao-an, to give some context: we're facing out log4j 1.x due to all the CVEs popping up lately. While those were for 2.x, there's others for 1.x, too. Those don't affect the usage in Debezium, and this used to be a test-only dependency. So there was no _actual_ problem, but then again solely having that dependency can raise questions for folks who look at CVE reporting tools and see some "usage" of a CVE-ed dependency. Hence the decision to leave it behind entirely.

gunnarmorling:
Btw. note that actual logging at runtime in Kafka Connect still is done via log4j 1.x. This is only going to change once the upstream Kafka community has moved off of log4j 1.x. But from a Debezium perspective, we have gotten rid of it with this.

shichao-an:
@gunnarmorling Got it. Thanks for the updates!

PR #75, opened 2022-03-01
DBZ-4324: Remove vtctld from sample config

y5w:
Follow up from https://github.com/debezium/debezium-connector-vitess/pull/50

Now that we don't depend on vtctld, remove it from the sample config file


shichao-an (approved):
LGTM

gunnarmorling:
Hey, it doesn't matter that much for this particular change, but in general let's avoid attributing commits to issues which have been shipped already (i.e. status in Jira is "Closed"). It's bending history a bit, as one might (wrongly) assume this change would be part of 1.8.0.Beta1 (the fix version of DBZ-4324). Instead, let's create a follow-up issue for such change in this situation, which can be linked to the original one to clarify the relationship. Thanks!

y5w:
> Hey, it doesn't matter that much for this particular change, but in general let's avoid attributing commits to issues which have been shipped already (i.e. status in Jira is "Closed"). It's bending history a bit, as one might (wrongly) assume this change would be part of 1.8.0.Beta1 (the fix version of DBZ-4324). Instead, let's create a follow-up issue for such change in this situation, which can be linked to the original one to clarify the relationship. Thanks!

@gunnarmorling Ack, will follow the suggestion, Thanks!

PR #76, opened 2022-03-08
DBZ-4673: Filter out empty transaction events

y5w:
Resolves https://issues.redhat.com/browse/DBZ-4673

The TL;DR here is when we start a vstream for a keyspace (i.e., a database in mysql terminology), Vitess will only produce meaningful events for that database. 

All the operational writes to other internal databases will be redacted. As an example, if our database name is "ks" and we do a write to "test" database on the same mysqld, what we will see from the Vitess is:
```
{
  "events": [
    {
      "type": "BEGIN",
      "timestamp": "1646712569",
      "currentTime": "1646712569872831735",
      "keyspace": "ks",
      "shard": "0"
    },
    {
      "type": "VGTID",
      "vgtid": {
        "shardGtids": [
          {
            "keyspace": "ks",
            "shard": "0",
            "gtid": "MySQL56/688caaaf-94c5-4243-a00f-0f19a39a7a37:1-11883129"
          }
        ]
      },
      "keyspace": "ks",
      "shard": "0"
    },
    {
      "type": "COMMIT",
      "timestamp": "1646712569",
      "currentTime": "1646712569872865890",
      "keyspace": "ks",
      "shard": "0"
    }
  ]
}
```

You can see we have the gtid created but there is no meaningful things to be stream'd out. This pattern is common for some Vitess operations during schema migration, reshard, etc.

As a result, we can skip these events directly.

Note: I didn't add test case for this change mostly for two reasons:
- the change is relatively straightforward 
- the test setup is doing insert through vtgate, which only exposes the default database (i.e., it only expose `ks` database but not `test` database in our example)


y5w:
>That may result in a inconsistent way of processing events or any combination of events that we do not know about. [BEGIN, VGTID, END] is only one of them. There is also [BEGIN, END], etc.

That's why I'm explicitly checking `[BEGIN, VGTID, END]` - I don't want to mess around with other use cases yet.

> Also think about the use cases where we actually need this events? 

These events only have GTID, I couldn't think of a use case other than checking data quality (i.e., make sure we don't lose any events) - but even in that case, I think it won't help if there is a failover at the source db.

> We don't really want to blindly inject our own use cases or make assumptions in a open source project.

I don't think this is our own use case only. For example, we would see reshard operation from Vitess itself generates these kind of `[BEGIN, VGTID, END] ` events as well.

Good call on the flag, I will make it configurable

src/main/java/io/debezium/connector/vitess/connection/VitessReplicationConnection.java
```diff
@@ -99,6 +99,12 @@ public void onNext(Vtgate.VStreamResponse response) {
                     LOGGER.debug("VEvent: {}", vEvent);
                 }
 
+                // If we receive an empty transaction, there is nothing we can stream out
+                // to the downstream. Therefore, skip them directly here instead.
+                if (isEmptyTransaction(response)) {
```
shichao-an, line 104: Skipping events like this look a very coarse handling to me. It conflicts with existing logic on event processing and skipping, which is mostly done within `messageDecoder.processMessage`. See https://github.com/debezium/debezium-connector-vitess/pull/73/files for how [BEGIN, COMMIT] events were handled. I highly suggestwe  can handle this case similarly to avoid trapdoor.

src/main/java/io/debezium/connector/vitess/connection/VitessReplicationConnection.java
```diff
@@ -168,6 +174,15 @@ private int getNumOfRowEvents(Vtgate.VStreamResponse response) {
                 }
                 return num;
             }
+
+            // If we write to an internal database in Vitess for operation purpose
+            // Vitess will stream out an empty event with [BEGIN, VGTID, COMMIT].
+            private boolean isEmptyTransaction(Vtgate.VStreamResponse response) {
```
shichao-an, line 180: As I mentioned in the PR comment, consider making it configurable and log the skipped events.

shichao-an:
> These events only have GTID, I couldn't think of a use case other than checking data quality (i.e., make sure we don't lose any events) - but even in that case, I think it won't help if there is a failover at the source db.

From technical perspective, think of replication use case where we actually need to stream out gtids to the downstream so they can keep in sync (like VReplication).
From a non-technical perspective, think of users who want the [BEGIN, VGTID, COMMIT] caused by heartbeats because they want the heartbeats in topic.

y5w:
Close this PR for now because it might introduce a failure mode where a database does not have any real traffic, and binlog get purged out by retention policy. Without these "operational" events, the offset might point to the position that is not available in binlog anymore. We decided to do the filtering at downstream instead.

PR #78, opened 2022-03-20
DBZ-4886 Support maxInboundMessageSize config

y5w:
Resolves https://issues.redhat.com/browse/DBZ-4886

### DIFF
--- a/src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
+++ b/src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
@@ -133,6 +133,14 @@ public class VitessConnectorConfig extends RelationalDatabaseConnectorConfig {
             .withDescription("Specify a comma-separated list of gRPC headers." +
                     " Defaults to empty");
 
+    public static final Field GRPC_MAX_INBOUND_MESSAGE_SIZE = Field.create(VITESS_CONFIG_GROUP_PREFIX + "grpc.max_inbound_message_size")
+            .withDisplayName("VStream gRPC maxInboundMessageSize")
+            .withType(Type.INT)
+            .withWidth(Width.SHORT)
+            .withDefault(0)
+            .withImportance(ConfigDef.Importance.MEDIUM)
+            .withDescription("Specify the maximum message size allowed to be received on the channel.");
+
     public static final Field INCLUDE_UNKNOWN_DATATYPES = Field.create("include.unknown.datatypes")
             .withDisplayName("Include unknown datatypes")
             .withType(Type.BOOLEAN)
@@ -159,6 +167,7 @@ public class VitessConnectorConfig extends RelationalDatabaseConnectorConfig {
                     STOP_ON_RESHARD_FLAG,
                     KEEPALIVE_INTERVAL_MS,
                     GRPC_HEADERS,
+                    GRPC_MAX_INBOUND_MESSAGE_SIZE,
                     BINARY_HANDLING_MODE)
             .events(INCLUDE_UNKNOWN_DATATYPES)
             .excluding(SCHEMA_EXCLUDE_LIST, SCHEMA_INCLUDE_LIST)
@@ -253,6 +262,10 @@ public class VitessConnectorConfig extends RelationalDatabaseConnectorConfig {
         }
 
         return Collections.unmodifiableMap(grpcHeadersMap);
+    }
+
+    public int getGrpcMaxInboundMessageSize() {
+        return getConfig().getInteger(GRPC_MAX_INBOUND_MESSAGE_SIZE);
     }
 
     public boolean includeUnknownDatatypes() {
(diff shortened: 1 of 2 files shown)


src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
```diff
@@ -133,6 +133,14 @@
             .withDescription("Specify a comma-separated list of gRPC headers." +
                     " Defaults to empty");
 
+    public static final Field GRPC_MAX_INBOUND_MESSAGE_SIZE = Field.create(VITESS_CONFIG_GROUP_PREFIX + "grpc.max_inbound_message_size")
+            .withDisplayName("VStream gRPC maxInboundMessageSize")
+            .withType(Type.INT)
+            .withWidth(Width.SHORT)
+            .withDefault(0)
```
shichao-an, line 140: Default is 4194304 (4 MiB).
https://grpc.github.io/grpc-java/javadoc/io/grpc/ManagedChannelBuilder.html#maxInboundMessageSize-int-

src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
```diff
@@ -133,6 +133,14 @@
...
+            .withImportance(ConfigDef.Importance.MEDIUM)
+            .withDescription("Specify the maximum message size allowed to be received on the channel.");
```
shichao-an, line 142:
```suggestion
            .withDescription("Specify the maximum message size in bytes allowed to be received on the channel.");
```

src/main/java/io/debezium/connector/vitess/connection/VitessReplicationConnection.java
```diff
@@ -204,12 +204,14 @@ private int getNumOfRowEvents(Vtgate.VStreamResponse response) {
...
-    private ManagedChannel newChannel(String vtgateHost, int vtgatePort) {
-        ManagedChannel channel = ManagedChannelBuilder.forAddress(vtgateHost, vtgatePort)
+    private ManagedChannel newChannel(String vtgateHost, int vtgatePort, int maxInboundMessageSize) {
+        ManagedChannelBuilder channelBuilder = ManagedChannelBuilder.forAddress(vtgateHost, vtgatePort)
                 .usePlaintext()
-                .keepAliveTime(config.getKeepaliveInterval().toMillis(), TimeUnit.MILLISECONDS)
-                .build();
-        return channel;
+                .keepAliveTime(config.getKeepaliveInterval().toMillis(), TimeUnit.MILLISECONDS);
+        if (maxInboundMessageSize != 0) {
```
shichao-an, line 211: Since you already pass maxInboundMessageSize as an argument to newChannel, you might want to this value to always take effect. Just set the default value to 4 MiB and pass the value in as is.
 
0 is not a valid value anyway and the builder will check the argument and throw IllegalArgumentException

shichao-an (commented):
Thanks for the quick fix. Left some comments.

src/main/java/io/debezium/connector/vitess/connection/VitessReplicationConnection.java
y5w: The reason I use 0 as default is I want the connector to always grpc default. If they change the default value to something that is not 4MiB, the connector should follow that as well.

shichao-an: First, `newChannel(String vtgateHost, int vtgatePort, int maxInboundMessageSize)` makes people think `maxInboundMessageSize` is a required parameter for creating a new channel, but it's actually not.

Second, using 0 as the default value makes people think it's a valid value, but it's actually not.

Third, it's fine if "they" change the default value to something that is not 4MiB, since we still keep track the value explicitly in our code and docs all the time. By pinning the default value, we are responsible for it rather making it a loose knob prone to changes in third parties that could affect production without being noticed.

On top of "third", it's also bad user experience. Imagine if you set it to 0 by default, the user may assume two possibilities:

1. The max message size is not set (i.e. unlimited), they think they don't need to do anything since they trust value this without further thinking until some day they encounter the ` RESOURCE_EXHAUSTED: gRPC message exceeds maximum size 4194304` error.
2. The max message size is set to some kind of default value (they need to look up for it, but they cannot find anything in Debezium code, and they need to dig further and finally find some clues in gRPC docs, but they are still not sure whether it's 4 MiB or not because of implementation details). 

Let's say gRPC default value is indeed changed 4 MiB from 8 MiB some time in the future (which I highly doubt this will happen), how will the users know what default value is used here in our code? They need to know the 1) the first version of gRPC library that changes this value from 4 MiB to 8 MiB 2) version of gRPC library used in our code 3) A mapping of Debezium release versions and gRPC library versions.

In short, it's always better to be explicit than implicit. Let's not shoot ourselves in the foot.

y5w: I can see that interface is confusing but it is a `private` function and only being used by the class, which might be OK in this case. 

grpc config does not feel like something as a debezium user should care in this case, unless they ran into some weird issues that requires some customization to work around. I think _most_ of the developers might assume debezium follows grpc default settings. My intention was providing a way to allow the "overrides" instead of a source of truth.

I don't feel strong either way, updated the default to 4MiB and removed the 0 check in the code

shichao-an (approved):

src/main/java/io/debezium/connector/vitess/connection/VitessReplicationConnection.java
shichao-an: For optional arguments, I prefer simply get it from config, just like `config.getKeepaliveInterval()`.

PR #79, opened 2022-04-28
DBZ-5063 Fix inconsistent transaction id

shichao-an:
# Symptom
We found that the following events (raw output using grpcurl from VStream gRPC endpoint) immediately following some row events could cause the connector to output transaction metadata with the wrong GTID.

This is because the transaction id is cached in VStreamOutputMessageDecoder. Whenever there is a normal response that contains `{BEGIN, VGTID and COMMIT}`, the cache will be updated the used according to send out transactional messages.
However, this cache is retained even after the normal response has been processed. If the next time some partial transaction responses like below are received, it will cause these BEGIN and COMMIT events to have the old transaction id from that previously processed response.

```
{  // response 1
  "events": [
    {
      "type": "BEGIN",
      "timestamp": "1651088385",
      "currentTime": "1651099396717227695",
      "keyspace": "ks1",
      "shard": "0"
    },
    {
      "type": "VERSION",
      "timestamp": "1651088385",
      "currentTime": "1651099396720350986",
      "keyspace": "ks1",
      "shard": "0"
    }
  ]
}
{  // response 2
  "events": [
    {
      "type": "VGTID",
      "vgtid": {
        "shardGtids": [
          {
            "keyspace": "ks1",
            "shard": "0",
            "gtid": "MySQL56/0b978551-8c72-4183-9d07-9383140bb3a9:1-149087591"
          }
        ]
      },
      "keyspace": "ks1",
      "shard": "0"
    },
    {
      "type": "COMMIT",
      "timestamp": "1651088385",
      "currentTime": "1651099396720376470",
      "keyspace": "ks1",
      "shard": "0"
    }
  ]
}
```
# Cause Analysis

Vitess gRPC responses

As seen in the [README page](https://github.com/debezium/debezium-connector-vitess#managing-offset), a Vitess transaction could be split into multiple gRPC responses. The current code only handles multi-responses at offset level, meaning managing offset in both memory via VitessOffsetContext as well as emitting offset as part of the "source" field in the output records. However, the transaction IDs are not handled correctly. If a transaction is split into multiple responses, such as this [BEGIN, ROW1, ROW2] and [ROW2, VGTID, COMMIT], these events will be sent out using the previously cached transaction ID, which will cause the serious issues in the downstream consumers that rely on transaction ID to do transaction-level aggregation. 

The reason why a transaction can be received as multiple gRPC responses is that VStream will split large transactions into many chunks performance reasons:
* https://github.com/vitessio/vitess/blob/3d4f40041e34e781b4bf94b77ef70877b9f90c2d/go/vt/vttablet/tabletserver/vstreamer/packet_size.go#L27
* https://github.com/vitessio/vitess/blob/3d4f40041e34e781b4bf94b77ef70877b9f90c2d/go/vt/vttablet/tabletserver/vstreamer/vstreamer.go#L254

# Fix

To handle split transactions, we need to buffer events until we receive a complete transaction. When we receive a complete transaction, we can get the VGTID as tranasction id and process every single event in between using this transaction id. This will guarantee correct transaction id is attached to all events that belong to that particular transaction.

There might be other unpredicted behaviors or bugs from Vitess VStream, e.g. duplicate responses, unpaired BEGIN/COMMIT events. Therefore, to make connector code robust, we catch these cases and throw an error to the error handler.

JIRA ticket: https://issues.redhat.com/browse/DBZ-5063

### DIFF
--- a/src/main/java/io/debezium/connector/vitess/connection/VStreamOutputMessageDecoder.java
+++ b/src/main/java/io/debezium/connector/vitess/connection/VStreamOutputMessageDecoder.java
@@ -78,7 +78,7 @@ public class VStreamOutputMessageDecoder implements MessageDecoder {
             case VERSION:
                 break;
             default:
-                LOGGER.warn("VEventType {} skipped, not processing.", vEventType);
+                LOGGER.info("VEventType {} skipped, not processing.", vEventType);
         }
     }
 
@@ -107,10 +107,9 @@ public class VStreamOutputMessageDecoder implements MessageDecoder {
     private void handleBeginMessage(Binlogdata.VEvent vEvent, ReplicationMessageProcessor processor, Vgtid newVgtid)
             throws InterruptedException {
         this.commitTimestamp = Instant.ofEpochSecond(vEvent.getTimestamp());
-        // Use the entire VGTID as transaction id
-        if (newVgtid != null) {
-            this.transactionId = newVgtid.toString();
-        }
+        // Use the entire VGTID as transaction id.
+        // If newVgtid is null, reset transactionId to null.
+        this.transactionId = newVgtid == null ? null : newVgtid.toString();
         // Transaction ID must not be null in TransactionalMessage.
         if (this.transactionId == null) {
             LOGGER.info("Skip processing BEGIN because no VGTID was received");
@@ -133,6 +132,8 @@ public class VStreamOutputMessageDecoder implements MessageDecoder {
         LOGGER.trace("Commit timestamp of commit transaction: {}", commitTimestamp);
         processor.process(
                 new TransactionalMessage(Operation.COMMIT, transactionId, commitTimestamp), newVgtid, false);
+        // Reset the transaction ID after COMMIT.
+        this.transactionId = null;
     }
 
     private void decodeRows(Binlogdata.VEvent vEvent, ReplicationMessageProcessor processor, Vgtid newVgtid, boolean isLastRowEventOfTransaction)
@@ -415,6 +416,11 @@ public class VStreamOutputMessageDecoder implements MessageDecoder {
     @VisibleForTesting
     void setTransactionId(String transactionId) {
         this.transactionId = transactionId;
+    }
+
+    @VisibleForTesting
+    String getTransactionId() {
+        return this.transactionId;
     }
 
     private static String validateColumnName(String columnName, String schemaName, String tableName) {
--- a/src/main/java/io/debezium/connector/vitess/connection/VStreamOutputMessageDecoder.java
+++ b/src/main/java/io/debezium/connector/vitess/connection/VStreamOutputMessageDecoder.java
@@ -417,11 +417,6 @@ public class VStreamOutputMessageDecoder implements MessageDecoder {
         this.transactionId = transactionId;
     }
 
-    @VisibleForTesting
-    String getTransactionId() {
-        return this.transactionId;
-    }
-
     private static String validateColumnName(String columnName, String schemaName, String tableName) {
         int length = columnName.length();
         if (length == 0) {
--- a/src/main/java/io/debezium/connector/vitess/connection/VStreamOutputMessageDecoder.java
+++ b/src/main/java/io/debezium/connector/vitess/connection/VStreamOutputMessageDecoder.java
@@ -108,8 +108,9 @@ public class VStreamOutputMessageDecoder implements MessageDecoder {
             throws InterruptedException {
         this.commitTimestamp = Instant.ofEpochSecond(vEvent.getTimestamp());
         // Use the entire VGTID as transaction id.
-        // If newVgtid is null, reset transactionId to null.
-        this.transactionId = newVgtid == null ? null : newVgtid.toString();
+        if (newVgtid != null) {
+            this.transactionId = newVgtid.toString();
+        }
         // Transaction ID must not be null in TransactionalMessage.
         if (this.transactionId == null) {
             LOGGER.info("Skip processing BEGIN because no VGTID was received");
@@ -132,8 +133,6 @@ public class VStreamOutputMessageDecoder implements MessageDecoder {
         LOGGER.trace("Commit timestamp of commit transaction: {}", commitTimestamp);
         processor.process(
                 new TransactionalMessage(Operation.COMMIT, transactionId, commitTimestamp), newVgtid, false);
-        // Reset the transaction ID after COMMIT.
-        this.transactionId = null;
     }
 
     private void decodeRows(Binlogdata.VEvent vEvent, ReplicationMessageProcessor processor, Vgtid newVgtid, boolean isLastRowEventOfTransaction)
--- a/src/main/java/io/debezium/connector/vitess/connection/VitessReplicationConnection.java
+++ b/src/main/java/io/debezium/connector/vitess/connection/VitessReplicationConnection.java
@@ -5,8 +5,9 @@
  */
 package io.debezium.connector.vitess.connection;
 
+import java.util.ArrayList;
 import java.util.Collections;
-import java.util.LinkedList;
+import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.concurrent.TimeUnit;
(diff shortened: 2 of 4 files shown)


shichao-an:
@gunnarmorling @jpechane Hi, do you know why this build failed ? https://github.com/debezium/debezium-connector-vitess/runs/6205483615?check_suite_focus=true

I couldn't figure out what this means.
```
Error:  Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.1:testCompile (default-testCompile) on project debezium-api: Fatal error compiling: error: release version 17 not supported -> [Help 1]
```

jpechane:
@shichao-an Thanks for poitning this out. It is related to our move to using OpenJDK 17 for builds.

@vjuranek Could you please take care of that? Also please check `debezium-connector-db2` repositorory too. Thanks

vjuranek:
Should be fixed by PR #80

jeffchao:
@shichao-an is it ok to skip these malformed messages? What if a malformed message is a legit transaction that was split into 2? I'm curious if this happens only for heartbeats being in the middle (900ms default) or could happen otherwise. Just want to make sure we don't drop legit transactions.

shichao-an:
> @shichao-an is it ok to skip these malformed messages? What if a malformed message is a legit transaction that was split into 2? I'm curious if this happens only for heartbeats being in the middle (900ms default) or could happen otherwise. Just want to make sure we don't drop legit transactions.

So in the example events above: `{BEGIN, VERSION}` and `{VGTID, COMMIT}`, if we assume they are in one response like `{BEGIN, VERSION, VGTID, COMMIT}`, we still skip VERSION event with the current code, meaning we treat it like `{BEGIN, VGTID, COMMIT}`, which is literally a upstream "heartbeat" event. Therefore we only skipped the transaction id (VGTID).

If the transaction contains row events but is malformed, like `{BEGIN, ROW, VGTID, COMMIT}` split into `{BEGIN, ROW}` and `{VGTID, COMMIT}`, then we will still send out row events, but these row events will not contain transaction id. If we don't add the fix in this PR, then all of these events will be sent but using the wrong VGTID (the previous normal events).

shichao-an:
@jeffchao I don't whether we will ever receive malformed transaction that contain row events. If it does, there is no easy fix. But we never seen that so far. But `{BEGIN, VERSION}` this one may be expected according to VStream source code https://git.corp.stripe.com/stripe-private-oss-forks/vitess/blob/master/go/vt/vttablet/tabletserver/vstreamer/vstreamer.go#L214-L217 
In other words, this PR assumes row events transaction will not be split in two or more parts, and only those non-essential events will get split due to the seemingly expected behavior in VStream. In this case we will not lose data or have inconsistent transactions for row events.

shichao-an:
FYI, I updated the PR with a better approach. Will follow up with detailed descriptions

src/main/java/io/debezium/connector/vitess/connection/VStreamOutputMessageDecoder.java
```diff
@@ -78,7 +78,7 @@ public void processMessage(Binlogdata.VEvent vEvent, ReplicationMessageProcessor
             case VERSION:
                 break;
             default:
-                LOGGER.warn("VEventType {} skipped, not processing.", vEventType);
+                LOGGER.info("VEventType {} skipped, not processing.", vEventType);
```
jeffchao, line 81: Minor thing - why change this to info? We might lose this message depending on log level.

shichao-stripe: Warning level is too high. We currently see a lot of SAVEPOINT in logs (500/s), so I had to change admin logger to change log level to "error" (above "warning") to disable this log.

By changing it to "info", we only need to override it to "warn", giving our warning log lines in this class an opportunity to show up in the log

src/main/java/io/debezium/connector/vitess/connection/VitessReplicationConnection.java
```diff
@@ -89,28 +91,86 @@ public void startStreaming(
...
+                            break;
+                    }
+                    bufferedEvents.add(event);
+                }
+
+                numResponses++;
+
+                // We only proceed when we receive a complete transaction after seeing both BEGIN and COMMIT events,
+                // OR if sendNow flag is true (meaning we should process buffered events immediately).
+                if ((!beginEventSeen || !commitEventSeen) && !sendNow) {
```
y5w, line 152: This seems to be making the assumption that transactions can be sent via multiple parts, and we are being defensive here. I wonder for cases like: `begin1, row1, commit1, begin2`, we would send them out, but the second begin event would still send out with wrong gtid (it will use the gtid from txn1 in this case iiuc) - basically VEvents are ordered but can be split arbitrarily.

Do we need to break the for-loop on on line 106 on commit/ddl/other events and rotate `bufferedEvents`?

shichao-stripe: Yeah. We are not just being defensive. Transactions CAN be split into multiple parts if each part exceeds 250KB (by default), which is the case we encountered.
However, they will not be arbitrarily split. They will only split between BEGIN and COMMIT, therefore `BEGIN1, ROW1, COMMIT1, BEGIN2` should exist as per VStream source code, because right upon COMMIT1, it must send out in its own response.

y5w: > Transactions CAN be split into multiple parts if each part exceeds 250KB (by default), which is the case we encountered.

For my own understanding, is this limitation on Vitess or MySQL?

src/test/java/io/debezium/connector/vitess/VitessConnectorIT.java
```diff
@@ -380,6 +386,64 @@ public void shouldUsePrevVgtidAsOffsetWhenNoVgtidInGrpcResponse() throws Excepti
...
+            for (int i = 1; i <= expectedRecordsCount; i++) {
+                SourceRecord record = assertRecordInserted(TestHelper.TEST_UNSHARDED_KEYSPACE + ".numeric_table", TestHelper.PK_FIELD);
+                final Struct txn = ((Struct) record.value()).getStruct("transaction");
+                String txId = txn.getString("id");
+                assertThat(txId).isNotNull();
+                assertThat(txId).isEqualTo(expectedTxId);
+                Vgtid actualVgtid = Vgtid.of(txId);
+                assertThat(actualVgtid).isNotEqualTo(baseVgtid);
+            }
+            assertRecordEnd(expectedTxId, expectedRecordsCount);
```
jeffchao, line 439: Can we add another assertion to ensure the vgtid legitimately changes on the commit boundary? A regular (non-batch) transaction would be fine.

src/main/java/io/debezium/connector/vitess/connection/VitessReplicationConnection.java
shichao-an: It's in Vitess: https://github.com/vitessio/vitess/blob/main/go/vt/vttablet/tabletserver/vstreamer/packet_size.go#L27

src/main/java/io/debezium/connector/vitess/connection/VitessReplicationConnection.java
```diff
@@ -89,28 +91,86 @@ public void startStreaming(
...
+                boolean sendNow = false;
+                for (VEvent event : response.getEventsList()) {
+                    LOGGER.debug("VEvent: {}", event);
+                    switch (event.getType()) {
+                        case ROW:
+                            numOfRowEvents++;
+                            break;
+                        case VGTID:
+                            // We always use the latest VGTID if any.
+                            newVgtid = Vgtid.of(event.getVgtid());
```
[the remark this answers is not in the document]
shichao-an: The previous code is literally last-write-win, though in the same response (more than one vgtid). I can add a warning here if we see multiple vgtid.

src/main/java/io/debezium/connector/vitess/connection/VitessReplicationConnection.java
```diff
@@ -89,28 +91,86 @@ public void startStreaming(
...
+                            if (!beginEventSeen) {
+                                LOGGER.error("Receive COMMIT event before receiving BEGIN event");
+                            }
+                            if (commitEventSeen) {
+                                LOGGER.error("Received duplicate COMMIT events");
+                            }
+                            commitEventSeen = true;
+                            break;
+                        case DDL:
+                        case OTHER:
```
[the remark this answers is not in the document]
shichao-an: We just process them (send them to) to VStreamOutputMessageDecoder.processMessage, then in tje processor (VitessStreamingChangeEventSource.newReplicationMessageProcessor) it will merely rotate the vgtid for offset. These two events will be not be send as SourceRecords to Kafka.

src/main/java/io/debezium/connector/vitess/connection/VitessReplicationConnection.java
```diff
@@ -89,28 +91,86 @@ public void startStreaming(
...
+                if ((!beginEventSeen || !commitEventSeen) && !sendNow) {
+                    return;
                 }
 
-                Vgtid newVgtid = getVgtid(response);
-                int numOfRowEvents = getNumOfRowEvents(response);
+                if (numResponses > 1) {
+                    LOGGER.info("Processing multi-response transaction: number of responses is {}", numResponses);
+                }
+                if (newVgtid == null) {
```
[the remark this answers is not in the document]
shichao-an: Yes, sometimes there is no vgtid and it's expected, for example in the case of [BEGIN, COMMIT] as we see. In this case, we want to simply skip such events, because they cannot do anything (neither rotate vgtid or sending new recorsd as a transaction). 

BEGIN/END messages with null transaction id cannot be sent. However, with current untouched code, if we process them, it will make [BEGIN, COMMIT] to use the vgtid of the previous transaction, which is what we want to avoid. Therefore I'd better just skip it here, wdyt?

jeffchao (approved):
Great. Looking good after review. We'll get this into 1.9.3 but try nightly SNAPSHOT in case there's more issues we can address before it reaches the broader community.

y5w (approved):

PR #83, opened 2022-06-08
DBZ-5170 Mysql Commit Timestamp

harveyyue:
https://issues.redhat.com/browse/DBZ-5170

### DIFF
--- a/src/main/java/io/debezium/connector/vitess/VitessStreamingChangeEventSource.java
+++ b/src/main/java/io/debezium/connector/vitess/VitessStreamingChangeEventSource.java
@@ -96,11 +96,11 @@ public class VitessStreamingChangeEventSource implements StreamingChangeEventSou
                 offsetContext.rotateVgtid(newVgtid, message.getCommitTime());
                 if (message.getOperation() == ReplicationMessage.Operation.BEGIN) {
                     // send to transaction topic
-                    dispatcher.dispatchTransactionStartedEvent(partition, message.getTransactionId(), offsetContext);
+                    dispatcher.dispatchTransactionStartedEvent(partition, message.getTransactionId(), offsetContext, message.getCommitTime());
                 }
                 else if (message.getOperation() == ReplicationMessage.Operation.COMMIT) {
                     // send to transaction topic
-                    dispatcher.dispatchTransactionCommittedEvent(partition, offsetContext);
+                    dispatcher.dispatchTransactionCommittedEvent(partition, offsetContext, message.getCommitTime());
                 }
                 return;
             }


src/main/java/io/debezium/connector/vitess/VitessStreamingChangeEventSource.java
```diff
@@ -96,11 +96,11 @@ private ReplicationMessageProcessor newReplicationMessageProcessor(VitessPartiti
                 offsetContext.rotateVgtid(newVgtid, message.getCommitTime());
                 if (message.getOperation() == ReplicationMessage.Operation.BEGIN) {
                     // send to transaction topic
-                    dispatcher.dispatchTransactionStartedEvent(partition, message.getTransactionId(), offsetContext);
+                    dispatcher.dispatchTransactionStartedEvent(partition, message.getTransactionId(), offsetContext, message.getCommitTime());
```
shichao-an, line 99: Just for my own understanding. does this mean the transaction metadata messages (BEGIN/END) will have an extra field "commit_ts"? Something like this?
```
  "payload": {
    "status": "END",
    "id": "[{\"keyspace\":\"test_unsharded_keyspace\",\"shard\":\"0\",\"gtid\":\"MySQL56/e03ece6c-4c04-11ec-8e20-0242ac110004:1-37\"}]",
    "event_count": 1,
    "data_collections": [
      {
        "data_collection": "test_unsharded_keyspace.my_seq",
        "event_count": 1
      }
     "commit_ts": 1234
    ]
```

jingtaozhang18: Hi @jeffchao and @shichao-an , extracting the accurate **commit** time from binlog (at least the order of commit time is the same as in binlog) is a very important feature to build a historical table. In the historical table, we will use the commit time of records to identify when the records start to become visible, as shown below:

ID 1, Version 1, commit at 2022.06.15 09:00:00
ID 1, Version 2, commit at 2022.06.15 09:10:32

The version 1 of ID 1 is visible from 2022.06.15 09:00:00 to 2022.06.15 09:10:32
Ther version 2 of ID 1 is visible from 2022.06.15 09:10:32 to now.

Now, Debezium extract the time from binlog is the statement starts executing time, which is not accurate and maybe unordered, refer to [discussion](https://github.com/debezium/debezium/pull/3555)

I also want to hear your concerns and questions.

shichao-an: @harveyyue @jingtaozhang18 Thanks for the explanation!

That makes sense. In Vitess connector, this commit timestamp is originally retrieved from VEvent's timestamp field, which represents the binlog timestamp.

The only FUD (not related to your PR) is about the timestamp precision which is in second granularity, but this is an intrinsic limitation of MySQL, so there's nothing we can do about it I assume.

shichao-an (approved):

src/main/java/io/debezium/connector/vitess/VitessStreamingChangeEventSource.java
jeffchao: @jingtaozhang18 

I don't see any other way to get timestamp data other than what Vitess/MySQL provides without upstream changes.

That said, it sounds like you're trying to get total ordering (across transactions). I would be careful doing that using timestamps as it could be error-prone as you saw.

In particular, the correctness in ordering is only as good as the precision of the timestamps. Given Vitess/MySQL seconds-level granularity, you could potentially have collisions within a second and be forced either to have a non-deterministic ordering or to use some heuristic to do a second-level ordering (which is still error-prone).

Ordering by timestamp also presumes you'll need to buffer for some amount of time, which still leads to correctness vs freshness trade-offs: how long do you buffer to wait for an event to come? The longer you wait, the higher the chance for the event to arrive and you can correctly order, but then that means a longer delay for you to materialize events to your downstreams. The shorter you wait, the event may not arrive in time, which means you may end up missing transactions that arrive just a bit too late.

If you want total ordering, it's going to be better to use a single partition/parallelism in your data pipeline. However that runs the trade-off of throughput.

If you can relax your ordering constraint, for example ordering within a transaction (not across transactions), then you can still have arbitrary parallelism, but make sure you key your events by transaction ID so events with the same transaction ID go to the same partition for aggregation.

@harveyyue PR looks good. LGTM.

jeffchao (approved):

jingtaozhang18 (approved):

src/main/java/io/debezium/connector/vitess/VitessStreamingChangeEventSource.java
jingtaozhang18: @jeffchao 

What you said helped me a lot!

The timestamp precision just as @shichao-an said is an intrinsic limitation of MySQL. So it is not possible to sort by timestamp.

Now I am trying to use binlog index and binlog position to order transactions, which still has restrictions. Because binlog index and position is only in order in single mysql node. If this node is down, I can't switch a replica node to solve problem. And I also considered using GTID, but I can't distinguish the order when master-slave switchs.

Now I don't consider the order within a transaction. I think changes occur at the granularity of transactions. All changes in a transaction take effective at the same time when the transaction commits. Additionally I will only keep the last state of an entity within a transaction during the offline processing.

Finally, Thanks the contribution of @harveyyue

jeffchao: No problem. Glad it was helpful!

PR #87, opened 2022-07-07
DBZ-5368: Vitess Connector needs new authentication mechanism to hide…

HenryCaiHaiying:
… the plain password

The current debezium vitess connector is using plain username/password as connector's config parameters.  The plain password is sent through a REST call over the wire from client to Kafka Conenctor instance and is persisted as plain text in Connect's Kafka config topic.  This is not a secure way dealing with password and disallowed in My company's environment.

Instead we should use the client_auth file from the connector instance's local disk to avoid sending password over the wire.  The client auth file contains username/password for vitess vtgate authentication, the file is usually stored local on connector instance host.  This the vitess and grpc's recommended way for client authentication.  See example at: https://github.com/vitessio/vitess/blob/main/examples/local/grpc_static_client_auth.json

The example client_auth file is a json file having the following content:

    {
      "Username": "vitess",
      "Password": "vitess_password"
    }

A new config parameter database.auth_file is introduced for VitessConnectorConfig to specify the location of the client_auth file.  When this config parameter is present, the connector will read the username and password from the client_auth file instead of from database.username and database.password config parameters.  Both uppercase and lower case Username/Password field names are recognized in the client auth json file.

Added unit tests and integration tests to cover various positive/negative test cases (valid/invalid/empty auth file)

Also increased the maven.docker plugin's wait time to 240 seconds for vitess docker instance to be fully up (it's a big docker container starting multiple processes inside)


src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
```diff
@@ -224,12 +241,52 @@ public int getVtgatePort() {
         return getConfig().getInteger(VTGATE_PORT);
     }
 
+    private void checkVtgateAuthFile() {
+        if (_username != null || _password != null) {
+            return;
+        }
+        String authFile = getVtgateAuthFile();
+        if (authFile != null && !authFile.isEmpty()) {
+            JSONParser parser = new JSONParser();
```
shichao-an, line 250: Can we use Jackson? I remember we were trying to remove other JSON dependencies such as Gson from Debezium BOM. cc @gunnarmorling

src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
```diff
@@ -68,6 +73,14 @@ public class VitessConnectorConfig extends RelationalDatabaseConnectorConfig {
             .withImportance(ConfigDef.Importance.HIGH)
             .withDescription("Password of the user to be used when connecting the Vitess VTGate gRPC server.");
 
+    public static final Field VTGATE_AUTH_FILE = Field.create(DATABASE_CONFIG_PREFIX + "auth_file")
+            .withDisplayName("Client Auth File")
+            .withType(Type.STRING)
+            .withWidth(Width.LONG)
+            .withImportance(ConfigDef.Importance.HIGH)
+            .withDescription("Client auth file to be used when connecting to the VTGate server. " +
+                    "Sample: https://github.com/vitessio/vitess/blob/main/examples/local/grpc_static_client_auth.json");
```
shichao-an, line 82: Instead of linking an example here, I suggest skipping this here but explicitly providing an example in the documentation like this https://debezium.io/documentation/reference/stable/connectors/vitess.html#vitess-property-message-key-columns

The documentation source is in https://github.com/debezium/debezium/blob/main/documentation/modules/ROOT/pages/connectors/vitess.adoc

src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
```diff
@@ -184,6 +198,9 @@ public static ConfigDef configDef() {
         return CONFIG_DEFINITION.configDef();
     }
 
+    private String _username;
+    private String _password;
```
shichao-an, line 202: How about `usernameFromAuthFile` and `passwordFromAuthFile`. I don't think underscore prefix is a common Java idiom.

src/test/java/io/debezium/connector/vitess/TestHelper.java
```diff
@@ -49,14 +49,15 @@ public class TestHelper {
     // Use the same username and password for vtgate and vtctld
     private static final String USERNAME = "vitess";
     private static final String PASSWORD = "vitess_password";
+    private static final String PASSWORD_FILE = "vitess_password_file.json";
```
shichao-an, line 52: Maybe name it "vitess_auth_file.json" since it contains both username and password?

src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
```diff
@@ -224,12 +241,52 @@ public int getVtgatePort() {
         return getConfig().getInteger(VTGATE_PORT);
     }
 
+    private void checkVtgateAuthFile() {
```
shichao-an, line 244: Maybe name it something like `loadVtgateAuthFile()` since you are actually loading the credential from file into memory.

src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
HenryCaiHaiying: What's the problem of JSON or GSON library?  I can convert the code to use Jackson's ObjectMapper, but I think the code will be more verbose.

src/test/java/io/debezium/connector/vitess/TestHelper.java
HenryCaiHaiying: Yes, I can rename the file name.  Initially I introduced the new config as 'database.password_file' so some of the test cases still use password_file.  Later I decide to use config name 'database.auth_file' but haven't updated the test file name yet.

src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
jeffchao: +1 try to avoid more than 1 json serialization library. This will also make graduating from incubation and into core easier.

src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
HenryCaiHaiying: I can rename the variable names.  Some java packages (and some companies) do use _ as the prefix for instance variables.

src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
HenryCaiHaiying: I can name the method: 'maybeLoadVtgateAuthFile' since it's conditional loading.

src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
jeffchao: The config table is nice. Consider adding a short text around https://debezium.io/documentation/reference/2.0/connectors/vitess.html#setting-up-vitess as well on the setup.

jeffchao:
@HenryCaiHaiying thanks for the contribution. Few comments. I'll circle back tomorrow (UTC -8).

src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
HenryCaiHaiying: OK, let me convert the code using Jackson's ObjectMapper

HenryCaiHaiying:
@jeffchao The JIRA ticket is here: https://issues.redhat.com/browse/DBZ-5368

jpechane:
Hi guys, please make sure we are not replicating https://debezium.io/blog/2019/12/13/externalized-secrets/

HenryCaiHaiying:
Closing PR since the issue can be resolved with https://debezium.io/blog/2019/12/13/externalized-secrets/

PR #91, opened 2022-08-04
DBZ-5382: Support multiple tasks in vitess connector

HenryCaiHaiying:
The current Vitess Connector can be launched in two modes:

1. Don't specify Vitess shard info, this will make the connector task to subscribe to ViTess VStream for all shards in the given KeySpace;
2. Specify a particular Vitess shard info, this will make the connector task to subscribe to VStream just for this shard;

In both modes, the connector will only launch one source task (VitessConnectorTask) to do the actual VStream processing.

For a bigger Vitess installation which has many shards, one vitess-dbzium connector with only one source task (mode 1) is not going to be able to sustains the event stream from Vitess.  For mode 2, the operation person needs to manually figure out all the shards in the vitess and manually launch many vitess-dbzium connectors (one connector for one Vitess shard).  This is very error prone and might end up with 1000 connectors to launch and manage.

In Kafka-Connect's deployment model, the recommendation is to launch one connector and have the connector do the service discovery of Vitess shards and launch appropriate number of source tasks to manage the traffic load from Vitess side.

Enable the multi-tasking support by specifying the connect config param tasks.max > 1 and :

1. Add a new config vitess.offset.storage.per.task to incidate that we
   want to partition the offset.storage (stored in kafka topic
   connect-offsets) and the partition key is the task id;  When this new
   config param is turned on, the offset storage is going to be changed
   from keyed by database.server name to task ids:

   Old offset storage:
```
        Key ["vitess-connector",{"server":"dbServer"}],
        Value: {"transaction_id":null,"vgtid":"[
            {\"keyspace\":\"byuser\",\"shard\":\"-4000\",\"gtid\":\"MySQL56/027c67a2-c0b0-11ec-8a34-0ed0087913a5:1-11418261\"},
            {\"keyspace\":\"byuser\",\"shard\":\"4000-8000\",\"gtid\":\"MySQL56/1221c6f4-c0b0-11ec-990f-0e78f7c0adfb:1-7301701\"},
            {\"keyspace\":\"byuser\",\"shard\":\"8000-c000\",\"gtid\":\"MySQL56/00b87c67-369a-11ec-8de9-12b6db74a3c3:1-28\"},
            {\"keyspace\":\"byuser\",\"shard\":\"c000-\",\"gtid\":\"MySQL56/037f120a-369a-11ec-b886-06529ac1b6b1:1-199888\"}]"}
```
   New offset storage partioned into two partitions (assuming the user specified tasks.max = 2):
```
        Key ["vitess-connector",{"server":"task0_2"}],
        Value: {"transaction_id":null,"vgtid":"[
            {\"keyspace\":\"byuser\",\"shard\":\"-4000\",\"gtid\":\"MySQL56/027c67a2-c0b0-11ec-8a34-0ed0087913a5:1-11418261\"},
            {\"keyspace\":\"byuser\",\"shard\":\"8000-c000\",\"gtid\":\"MySQL56/00b87c67-369a-11ec-8de9-12b6db74a3c3:1-28\"}]"}
        Key ["vitess-connector",{"server":"task1_2"}],
        Value: {"transaction_id":null,"vgtid":"[
            {\"keyspace\":\"byuser\",\"shard\":\"4000-8000\",\"gtid\":\"MySQL56/1221c6f4-c0b0-11ec-990f-0e78f7c0adfb:1-7301701\"},
            {\"keyspace\":\"byuser\",\"shard\":\"c000-\",\"gtid\":\"MySQL56/037f120a-369a-11ec-b886-06529ac1b6b1:1-199888\"}]"}
```
   The partition key is essentially the taskId appended by total number of tasks

2. A new method VitessConnector#getVitessShards is added to retrieve the
   current set of vitess shards for the configured keyspace.  This is
   achieved by parsing the query result from query: "SHOW VITESS_SHARDS"

3. VitessConnector#taskConfigs(int maxTasks) method is modified to break the space into multiple partitions and assign each partition to each task by returning a List of config maps (one map for each connector source task)
   e.g. for maxTasks=2 and current vitess shards = ["-4000","4000-8000","8000-c000","c000-"], the connector will break the space into two partitions and return a List with 2 elements for the Kafka Connect framework to launch two source tasks:
```
   [
      {
         "vitess.task.key": "task0_2",   // format: taskId_numOfTasks
         "vitess.task.shards": "-4000,8000-c000",   // format: comma separated vitess shard names
         ... other connector configs ...
      },
      {
         "vitess.task.key": "task1_2",
         "vitess.task.shards": "4000-8000,c000-",
         ... other connector configs ...
      }
   ]
```

4. When the source task (VitessConnectorTask) starts, it will find the "vitess.task.key" value (e.g. task0_2) assigned to the task, and will use this value (task0_2) to find the previous offsets stored in OffsetStorage.  If no previous offsets were found, VitessReplicationConnection#defaultVGtid() method will be invoked which will generate the default vgtid ("current") for the shards assigned to the task (the list of shards are passed in through "vitess.task.shard" param, in this case the value is "-4000,8000-c000").  VStream will be subscribed using this VGtid, When the VStream flows with the new Vgtids, the new VGtids will be persisted in the current task partition space (in this case: "task0_2").  When the tasks (or the whole connector runtime) restarts, "task0_2" might be assigned to a differen source task object, but this is fine since each source task will be assigned with a unique 'vitess.task.key" and each one will respectively retrieve the old VGtids stored in OffsetStorage for that partition key and continues.

5. The vitess shards to task assignment is a simple round-robin distribution (shard 0,2,4,... will be given to task 0, shards 1,3,5,... will go to task 1), in the future more advanced assignment algorithm can be introduced.  In the case when the number of vitess shards is less than the number of tasks, tasks.max will be reduced to the number of shards.  The connector is not expected to return the list with the exact tasks.max elements;

6. Some validations are added to make sure we only support multi-task mode (tasks.max > 1) when vitess.offset.storage.per.task=true

Note this PR does not address the automatic offset migration when the tasks.max is changing nor the initial migration when vitess.offset.storage.per.task is first set to true.  In both situations, the vitess connector will report no previous offsets were found in offset storage and will default VGtid to "current" for those shards.  Handling offset auto migration is not trivial since Kafka Connect framework only exposes reader API. Changing task scale factor is not going to be a frequent operation, and we recommend using an external script to directly update the Kafka offset topic (e.g. using kafkacat command as in the below example):
```
    a. sudo apt install kafkacat
    b. reading the current content of kafka connect offset topic (cdc-connect-offsets) (the below command will also print out the kafka key and partition number of the message):
          kafkacat -b localhost:9092 -C -t cdc-connect-offsets -f 'Topic %t [%p] at offset %o: Key %k, Value: %s\n'
    c. updating the content of kafka connect offset topic (this example is updating the partition key from old dbServerName to taskId_numTasks: "task0_1"):
          echo "[\"vitess-connector\",{\"server\":\"task0_1\"}]={"transaction_id":null,"vgtid":"<old-vgtid-from-previous-command"}" | kafkacat -b localhost:9092 -t cdc-connect-offsets -p 22 -K=
       note the above command also specify the kafka partition number 22 (retrieved from the previous read command) to make sure the script is using the same partition number as the Kafkak Connect java code
```

### DIFF
--- a/src/main/java/io/debezium/connector/vitess/VitessConnector.java
+++ b/src/main/java/io/debezium/connector/vitess/VitessConnector.java
@@ -235,16 +235,17 @@ public class VitessConnector extends RelationalBaseSourceConnector {
         LOGGER.info("Validating config: {}", config);
         Map<String, ConfigValue> results = config.validate(VitessConnectorConfig.ALL_FIELDS);
         Integer maxTasks = config.getInteger(VitessConnectorConfig.TASKS_MAX_CONFIG);
+        VitessConnectorConfig tempConnectorConfig = new VitessConnectorConfig(config);
         if (maxTasks != null && maxTasks > 1) {
-            if (!connectorConfig.offsetStoragePerTask()) {
+            if (!tempConnectorConfig.offsetStoragePerTask()) {
                 String configName = VitessConnectorConfig.OFFSET_STORAGE_PER_TASK.name();
                 results.computeIfAbsent(configName, k -> new ConfigValue(configName));
                 results.get(configName).addErrorMessage(String.format(
                         "%s needs to be enabled when %s > 1", configName, VitessConnectorConfig.TASKS_MAX_CONFIG));
             }
         }
-        if (connectorConfig.offsetStoragePerTask()) {
-            if (connectorConfig.getOffsetStorageTaskKeyGen() < 0) {
+        if (tempConnectorConfig.offsetStoragePerTask()) {
+            if (tempConnectorConfig.getOffsetStorageTaskKeyGen() < 0) {
                 String configName = VitessConnectorConfig.OFFSET_STORAGE_TASK_KEY_GEN.name();
                 results.computeIfAbsent(configName, k -> new ConfigValue(configName));
                 results.get(configName).addErrorMessage(String.format(
@@ -252,8 +253,8 @@ public class VitessConnector extends RelationalBaseSourceConnector {
                         configName, VitessConnectorConfig.OFFSET_STORAGE_PER_TASK.name()));
             }
         }
-        if (connectorConfig.getOffsetStorageTaskKeyGen() >= 0) {
-            if (connectorConfig.getPrevNumTasks() <= 0) {
+        if (tempConnectorConfig.getOffsetStorageTaskKeyGen() >= 0) {
+            if (tempConnectorConfig.getPrevNumTasks() <= 0) {
                 String configName = VitessConnectorConfig.PREV_NUM_TASKS.name();
                 results.computeIfAbsent(configName, k -> new ConfigValue(configName));
                 results.get(configName).addErrorMessage(String.format(
--- a/src/main/java/io/debezium/connector/vitess/VitessConnector.java
+++ b/src/main/java/io/debezium/connector/vitess/VitessConnector.java
@@ -73,7 +73,7 @@ public class VitessConnector extends RelationalBaseSourceConnector {
                     throw new IllegalArgumentException(String.format("No offset found for %s", par));
                 }
                 else {
-                    LOGGER.warn("No offset found for key key: {}", taskKey);
+                    LOGGER.warn("No offset found for task key: {}", taskKey);
                     continue;
                 }
             }
@@ -99,18 +99,22 @@ public class VitessConnector extends RelationalBaseSourceConnector {
         if (connectorConfig.offsetStoragePerTask()) {
             final int prevNumTasks = connectorConfig.getPrevNumTasks();
             final int gen = connectorConfig.getOffsetStorageTaskKeyGen();
-            Map<String, String> prevGtidsPerShard = gen > 0 ? getGtidPerShardFromStorage(prevNumTasks, gen - 1, true) : null;
-            LOGGER.info("Previous gtids Per shard: {}", prevGtidsPerShard);
             int tasks = Math.min(maxTasks, shards.size());
             LOGGER.info("There are {} vitess shards for maxTasks: {}, we will use {} tasks",
                     shards.size(), maxTasks, tasks);
             if (gen > 0 && tasks == prevNumTasks) {
                 throw new IllegalArgumentException(String.format(
-                        "Previous num.tasks: %s and current num.tasks: %s are the same",
+                        "Previous num.tasks: %s and current num.tasks: %s are the same. "
+                                + "Please choose different tasks.max or have different number of vitess shards "
+                                + "if you want to change the task parallelism.  "
+                                + "Otherwise please reset the offset.storage.task.key.gen config to its original value",
                         prevNumTasks, tasks));
             }
+            Map<String, String> prevGtidsPerShard = gen > 0 ? getGtidPerShardFromStorage(prevNumTasks, gen - 1, true) : null;
+            LOGGER.info("Previous gtids Per shard: {}", prevGtidsPerShard);
             if (prevGtidsPerShard != null && prevGtidsPerShard.size() != shards.size()) {
-                throw new IllegalArgumentException(String.format("Different number of shards between %s and %S",
+                throw new IllegalArgumentException(String.format(
+                        "Different number of shards between offset storage: %s and current vitess shards: %s. ",
                         prevGtidsPerShard, shards));
             }
             final String keyspace = connectorConfig.getKeyspace();
@@ -133,7 +137,7 @@ public class VitessConnector extends RelationalBaseSourceConnector {
                 List<String> taskShards = shardsPerTask.get(tid);
                 Map<String, String> taskProps = new HashMap<>(properties);
                 taskProps.put(VitessConnectorConfig.VITESS_TASK_KEY_CONFIG, getTaskKeyName(tid, tasks, gen));
-                taskProps.put(VitessConnectorConfig.VITESS_TASK_KEY_SHARDS_CONFIG, String.join(",", taskShards));
+                taskProps.put(VitessConnectorConfig.VITESS_TASK_SHARDS_CONFIG, String.join(",", taskShards));
                 List<Vgtid.ShardGtid> shardGtids = new ArrayList<>();
                 for (String shard : taskShards) {
                     String gtidStr = gtidsPerShard != null ? gtidsPerShard.get(shard) : null;
@@ -147,7 +151,7 @@ public class VitessConnector extends RelationalBaseSourceConnector {
                     }
                     shardGtids.add(new Vgtid.ShardGtid(keyspace, shard, gtidStr));
                 }
-                taskProps.put(VitessConnectorConfig.VITESS_KEY_KEY_VGTID_CONFIG, Vgtid.of(shardGtids).toString());
+                taskProps.put(VitessConnectorConfig.VITESS_TASK_VGTID_CONFIG, Vgtid.of(shardGtids).toString());
                 allTaskProps.add(taskProps);
             }
             LOGGER.info("taskConfigs are: {}", allTaskProps);
--- a/src/main/java/io/debezium/connector/vitess/VitessConnector.java
+++ b/src/main/java/io/debezium/connector/vitess/VitessConnector.java
@@ -76,7 +76,7 @@ public class VitessConnector extends RelationalBaseSourceConnector {
             for (int tid : shardsPerTask.keySet()) {
                 List<String> taskShards = shardsPerTask.get(tid);
                 Map<String, String> taskProps = new HashMap<>(properties);
-                taskProps.put(VitessConnectorConfig.VITESS_TASK_KEY_CONFIG, String.format("task%d_%d", tid, tasks));
+                taskProps.put(VitessConnectorConfig.VITESS_TASK_KEY_CONFIG, getTaskKeyName(tid, tasks));
                 taskProps.put(VitessConnectorConfig.VITESS_TASK_KEY_SHARDS_CONFIG, String.join(",", taskShards));
                 allTaskProps.add(taskProps);
             }
@@ -89,6 +89,10 @@ public class VitessConnector extends RelationalBaseSourceConnector {
             }
             return Collections.singletonList(properties);
         }
+    }
+
+    protected static final String getTaskKeyName(int tid, int numTasks) {
+        return String.format("task%d_%d", tid, numTasks);
     }
 
     @Override
--- a/src/main/java/io/debezium/connector/vitess/VitessConnector.java
+++ b/src/main/java/io/debezium/connector/vitess/VitessConnector.java
@@ -11,10 +11,13 @@ import java.util.Comparator;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 
 import org.apache.kafka.common.config.ConfigDef;
 import org.apache.kafka.common.config.ConfigValue;
 import org.apache.kafka.connect.connector.Task;
+import org.apache.kafka.connect.source.SourceConnectorContext;
+import org.apache.kafka.connect.storage.OffsetStorageReader;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -32,11 +35,14 @@ public class VitessConnector extends RelationalBaseSourceConnector {
     private static final Logger LOGGER = LoggerFactory.getLogger(VitessConnector.class);
 
     private Map<String, String> properties;
+    private VitessConnectorConfig connectorConfig;
 
     @Override
     public void start(Map<String, String> props) {
         LOGGER.info("Starting Vitess Connector");
         this.properties = Collections.unmodifiableMap(props);
+        this.connectorConfig = new VitessConnectorConfig(Configuration.from(properties));
+
     }
 
     @Override
@@ -44,25 +50,75 @@ public class VitessConnector extends RelationalBaseSourceConnector {
         return VitessConnectorTask.class;
     }
 
+    protected Map<String, String> getGtidPerShardFromStorage(int numTasks, int gen, boolean expectsOffset) {
+        // Note that in integration test, EmbeddedEngine didn't initialize SourceConnector with SourceConnectorContext
+        if (context == null
+                || !(context instanceof SourceConnectorContext) || context().offsetStorageReader() == null) {
+            LOGGER.warn("Context {} is not setup for the connector, this can happen in unit tests.", context);
+            return null;
+        }
+        final OffsetStorageReader offsetStorageReader = context().offsetStorageReader();
+        final Map<String, String> gtidsPerShard = new HashMap<>();
+        for (int i = 0; i < numTasks; i++) {
+            String taskKey = VitessConnector.getTaskKeyName(i, numTasks, gen);
+            VitessPartition par = new VitessPartition(connectorConfig.getLogicalName(), taskKey);
+            Map<String, Object> offset = offsetStorageReader.offset(par.getSourcePartition());
+            if (offset == null && gen == 0) {
+                LOGGER.info("No previous offset for partition: {}, fall back to only server key", par);
+                par = new VitessPartition(connectorConfig.getLogicalName(), null);
+                offset = offsetStorageReader.offset(par.getSourcePartition());
+            }
+            if (offset == null) {
+                if (expectsOffset) {
+                    throw new IllegalArgumentException(String.format("No offset found for %s", par));
+                }
+                else {
+                    LOGGER.warn("No offset found for key key: {}", taskKey);
+                    continue;
+                }
+            }
+            final String vgtidStr = (String) offset.get(SourceInfo.VGTID_KEY);
+            Objects.requireNonNull(vgtidStr, String.format("No vgtid from %s", offset));
+            List<Vgtid.ShardGtid> shardGtids = Vgtid.of(vgtidStr).getShardGtids();
+            for (Vgtid.ShardGtid shardGtid : shardGtids) {
+                gtidsPerShard.put(shardGtid.getShard(), shardGtid.getGtid());
+            }
+        }
+        return gtidsPerShard;
+    }
+
     @Override
     public List<Map<String, String>> taskConfigs(int maxTasks) {
         LOGGER.info("Calculating taskConfigs for {} tasks", maxTasks);
-        List<String> shards = null;
-        String storagePerTask = properties.get(VitessConnectorConfig.OFFSET_STORAGE_PER_TASK.name());
-        if (storagePerTask != null && storagePerTask.equalsIgnoreCase("true")) {
-            VitessConnectorConfig connectorConfig = new VitessConnectorConfig(Configuration.from(properties));
-            shards = getVitessShards(connectorConfig);
-        }
+        List<String> shards = connectorConfig.offsetStoragePerTask() ? getVitessShards(connectorConfig) : null;
         return taskConfigs(maxTasks, shards);
     }
 
     public List<Map<String, String>> taskConfigs(int maxTasks, List<String> shards) {
         LOGGER.info("Calculating taskConfigs for {} tasks and shards: {}", maxTasks, shards);
-        String storagePerTask = properties.get(VitessConnectorConfig.OFFSET_STORAGE_PER_TASK.name());
-        if (storagePerTask != null && storagePerTask.equalsIgnoreCase("true")) {
+        if (connectorConfig.offsetStoragePerTask()) {
+            final int prevNumTasks = connectorConfig.getPrevNumTasks();
+            final int gen = connectorConfig.getOffsetStorageTaskKeyGen();
+            Map<String, String> prevGtidsPerShard = gen > 0 ? getGtidPerShardFromStorage(prevNumTasks, gen - 1, true) : null;
+            LOGGER.info("Previous gtids Per shard: {}", prevGtidsPerShard);
             int tasks = Math.min(maxTasks, shards.size());
             LOGGER.info("There are {} vitess shards for maxTasks: {}, we will use {} tasks",
                     shards.size(), maxTasks, tasks);
+            if (gen > 0 && tasks == prevNumTasks) {
+                throw new IllegalArgumentException(String.format(
+                        "Previous num.tasks: %s and current num.tasks: %s are the same",
+                        prevNumTasks, tasks));
+            }
+            if (prevGtidsPerShard != null && prevGtidsPerShard.size() != shards.size()) {
+                throw new IllegalArgumentException(String.format("Different number of shards between %s and %S",
+                        prevGtidsPerShard, shards));
+            }
+            final String keyspace = connectorConfig.getKeyspace();
+            Map<String, String> gtidsPerShard = getGtidPerShardFromStorage(tasks, gen, false);
+            if (gtidsPerShard != null && gtidsPerShard.size() != shards.size()) {
+                LOGGER.warn("Some shards for the current generation {} are not persisted.  Expected shards: {}",
+                        gtidsPerShard.keySet(), shards);
+            }
             shards.sort(Comparator.naturalOrder());
             Map<Integer, List<String>> shardsPerTask = new HashMap<>();
             int taskId = 0;
@@ -76,8 +132,22 @@ public class VitessConnector extends RelationalBaseSourceConnector {
             for (int tid : shardsPerTask.keySet()) {
                 List<String> taskShards = shardsPerTask.get(tid);
                 Map<String, String> taskProps = new HashMap<>(properties);
-                taskProps.put(VitessConnectorConfig.VITESS_TASK_KEY_CONFIG, getTaskKeyName(tid, tasks));
+                taskProps.put(VitessConnectorConfig.VITESS_TASK_KEY_CONFIG, getTaskKeyName(tid, tasks, gen));
                 taskProps.put(VitessConnectorConfig.VITESS_TASK_KEY_SHARDS_CONFIG, String.join(",", taskShards));
+                List<Vgtid.ShardGtid> shardGtids = new ArrayList<>();
+                for (String shard : taskShards) {
+                    String gtidStr = gtidsPerShard != null ? gtidsPerShard.get(shard) : null;
+                    if (gtidStr == null) {
+                        LOGGER.warn("No current gtid found for shard: {}, fallback to previous gen", shard);
+                        gtidStr = prevGtidsPerShard != null ? prevGtidsPerShard.get(shard) : null;
+                    }
+                    if (gtidStr == null) {
+                        LOGGER.warn("No previous gtid found either for shard: {}, fallback to current", shard);
+                        gtidStr = Vgtid.CURRENT_GTID;
+                    }
+                    shardGtids.add(new Vgtid.ShardGtid(keyspace, shard, gtidStr));
+                }
+                taskProps.put(VitessConnectorConfig.VITESS_KEY_KEY_VGTID_CONFIG, Vgtid.of(shardGtids).toString());
                 allTaskProps.add(taskProps);
             }
             LOGGER.info("taskConfigs are: {}", allTaskProps);
@@ -91,8 +161,8 @@ public class VitessConnector extends RelationalBaseSourceConnector {
         }
     }
 
-    protected static final String getTaskKeyName(int tid, int numTasks) {
-        return String.format("task%d_%d", tid, numTasks);
+    protected static final String getTaskKeyName(int tid, int numTasks, int gen) {
+        return String.format("task%d_%d_%d", tid, numTasks, gen);
     }
 
     @Override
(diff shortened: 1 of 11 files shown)


src/main/java/io/debezium/connector/vitess/VitessConnector.java
```diff
@@ -41,10 +46,49 @@ public Class<? extends Task> taskClass() {
...
+                List<String> taskShards = shardsPerTask.computeIfAbsent(taskId, k -> new ArrayList<>());
+                taskShards.add(shard);
+                taskId = (taskId + 1) % tasks;
+            }
+            LOGGER.info("Shards task distribution: {}", shardsPerTask);
+            List<Map<String, String>> allTaskProps = new ArrayList<>();
+            for (int tid : shardsPerTask.keySet()) {
+                List<String> taskShards = shardsPerTask.get(tid);
+                Map<String, String> taskProps = new HashMap<>(properties);
+                taskProps.put(VitessConnectorConfig.VITESS_TASK_KEY_CONFIG, String.format("task%d_%d", tid, tasks));
```
shichao-an, line 79: What's purpose having a task key property with task count in the key?
I remember in Kafka Connect, connector task id are integer only right?
Is it for discriminating the shard setup under the certain task count only since once chosen it cannot be changed unless we do a migration.

src/main/java/io/debezium/connector/vitess/VitessConnector.java
```diff
@@ -41,10 +46,49 @@ public Class<? extends Task> taskClass() {
...
+            shards = getVitessShards(connectorConfig);
+        }
+        return taskConfigs(maxTasks, shards);
+    }
+
+    public List<Map<String, String>> taskConfigs(int maxTasks, List<String> shards) {
+        LOGGER.info("Calculating taskConfigs for {} tasks and shards: {}", maxTasks, shards);
+        String storagePerTask = properties.get(VitessConnectorConfig.OFFSET_STORAGE_PER_TASK.name());
+        if (storagePerTask != null && storagePerTask.equalsIgnoreCase("true")) {
+            int tasks = Math.min(maxTasks, shards.size());
```
shichao-an, line 63: How do you detect if maxTasks changes?
From your PR description you said you are not handling in this PR.
My thinking is that this would keep things silent behind the scene; If users didn't fully understand the implication and think vitess connector can handle that, they will have data loss if changing the max.tasks. Similarly to the case if they toggle `vitess.offset.storage.per.task` from true to false.

I'm thinking if we should catch that to log a warning or add a flag to decide whether to error out or just warn on migration related scenarios.

src/main/java/io/debezium/connector/vitess/VitessConnector.java
```diff
@@ -41,10 +46,49 @@ public Class<? extends Task> taskClass() {
 
     @Override
     public List<Map<String, String>> taskConfigs(int maxTasks) {
-        if (maxTasks > 1) {
-            throw new IllegalArgumentException("Only a single connector task may be started");
+        LOGGER.info("Calculating taskConfigs for {} tasks", maxTasks);
+        List<String> shards = null;
+        String storagePerTask = properties.get(VitessConnectorConfig.OFFSET_STORAGE_PER_TASK.name());
+        if (storagePerTask != null && storagePerTask.equalsIgnoreCase("true")) {
```
shichao-an, line 52: I saw you do this check twice: once here and another in the `taskConfigs` method below.
Is it cleaner to put everything about multi-task configs in that separate method after checking OFFSET_STORAGE_PER_TASK is enabled, and keep the original return value (i.e. `Collections.singletonList(properties)`) here?

Something like:

```
public List<Map<String, String>> taskConfigs(int maxTasks) {
  if (storagePerTask != null && storagePerTask.equalsIgnoreCase("true")) {
    // Put all shard listing and config generation in taskConfigsPerTask()
    return taskConfigsPerTask(...)
  }
  if (maxTasks > 1) {
    throw new IllegalArgumentException("Only a single connector task may be started");
  }
  return Collections.singletonList(properties);
}

```

This could save a lot of duplicate code and `else` branches

src/main/java/io/debezium/connector/vitess/VitessPartition.java
```diff
@@ -58,7 +63,16 @@ static class Provider implements Partition.Provider<VitessPartition> {
 
         @Override
         public Set<VitessPartition> getPartitions() {
-            return Collections.singleton(new VitessPartition(connectorConfig.getLogicalName()));
+            if (connectorConfig.offsetStoragePerTask()) {
+                String taskKey = connectorConfig.getVitessTaskKey();
+                if (taskKey == null) {
+                    throw new RuntimeException("No vitess.task.key in: {}" + connectorConfig.getConfig());
+                }
+                return Collections.singleton(new VitessPartition(taskKey));
```
shichao-an, line 71: It's a bit awkward to set create VitessPartition by populating server name (i.e. connector logical name in single-task setup) with the task key.
Shall we consider enhance VitessPartition to incorporate a meaningful field for task key. Something like https://github.com/debezium/debezium-connector-vitess/pull/61/files

src/main/java/io/debezium/connector/vitess/VitessConnector.java
HenryCaiHaiying: The reason I've put in both taskId and total_num_tasks was trying to handle the offset migration during task number change (or leave enough clues for other people to write a tool to handle the auto migration).

Kafka Connect's OffsetStorageReader class only has API to read a specific key, it doesn't have the API to list all the keys in the storage.  During a task number change (e.g. from tasks.max = 2 to tasks.max = 4), I need to figure out all the previous partition keys corresponding to tasks.max=2 generation, I was intending to have the user specify prev.tasks.max=2 in the config, so I will know all the keys for previous generation are: task0_2, task1_2.  This way I can read all the old vgtids for those key and reshuffle them for tasks.max=4.   And for this generation, I will generate task0_4, task1_4, task2_4, task2_4.  So the keys from two generation don't overlap.   If I don't have have the extra _4 in the key name, the partition keys between two generations will collide on task0 and task1.  We might enter into a situation some task didn't get the time to save new VGtid for task0 before it was killed accidentally.  On restart, it will retrieve the VGtid for task0 from offset storage which belong to the old generation.

But I am currently not handling offset auto migration nor do I use total_number_tasks in the task key, so I am open to either ways.  But I feel having total_number_tasks in the partition key name clearly shows which generation this partitioning belongs, it might be good for debugging as well.

src/main/java/io/debezium/connector/vitess/VitessConnector.java
HenryCaiHaiying: It's not easy to read the previous tasks.max value, by the time connector was called, the value in ConfigState has already updated.  We can try to detect the previous tasks.max by:
1. Fetch the keys in OffsetStorageReader, we don't have a list API in OffsetStorageReader so we need to guess read.  Currently the partition key in offsetStorageReader is in the form of task0_4, task1_4, so we should be able to try those keys to see whether the entry existed or not in offset storage.  If it doesn't, this is the first time connector is run with tasks.max=4.  This might still be fine if there is no previous run with other tasks.max value.  How do we know there is no previous offset history?  One way to blindly trying fetching partition keys: task0_1, task0_2, task0_3 for all the total_partition_number < 4, and left over partition keys in offset storage would generate a warning or error.  Of course this is assuming we would only scale up the number of tasks through the history
2. Have user explicitly specify the previous tasks.max

src/main/java/io/debezium/connector/vitess/VitessConnector.java
HenryCaiHaiying: I can do that.

src/main/java/io/debezium/connector/vitess/VitessPartition.java
HenryCaiHaiying: I see, but #61 PR hasn't been merged in yet.  Although we are only creating one VitessPartition object for the given source task, this VitessPartition object actually are working with multiple vitess shards (same way as the old VitessParition prior to this PR).  If you don't like the "server" in the partition key names, I can call it "shards" (note it's plurals, not the singular "shard" in PR 61).

src/main/java/io/debezium/connector/vitess/VitessConnector.java
HenryCaiHaiying: Actually the reason why I structured the code this way is I have a unit test which tests VitessConnector.taskConfigs()

blcksrx:
Its seems the CI fails because of first 2 commit messages.
Please use this doc to fix it:
https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/changing-a-commit-message#:~:text=On%20the%20command%20line%2C%20navigate,message%2C%20and%20save%20the%20commit.

HenryCaiHaiying:
What is the new content I need to put in for the previous 2 commit messages?
Or should I squash all the 4 commits into one commits?
From the console output, it seems it is complaining about the three extra files?

Run ! git log --format='format:%s' refs/remotes/origin/$BASE_REF.. | tail -n +2 | grep -Ev '^(DBZ-[[:digit:]]+)|(\[release\])|(\[jenkins-jobs\]) '

Remove three extra files

Merge branch 'DBZ-5382' of https://github.com/HenryCaiHaiying/debezium-connector-vitess into DBZ-5382
Error: Process completed with exit code 1.
    On Monday, August 8, 2022, 02:48:59 PM PDT, Hossein Torabi ***@***.***> wrote:  
 
 


Its seems the CI fails because of first 2 commit messages.
Please use this doc to fix it:
https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/changing-a-commit-message#:~:text=On%20the%20command%20line%2C%20navigate,message%2C%20and%20save%20the%20commit.

src/main/java/io/debezium/connector/vitess/VitessPartition.java
shichao-an: Yes. That PR was not merged, because there was no good use cases for it. 
But now this sounds a good fit.

> If you don't like the "server" in the partition key names, I can call it "shards" (note it's plurals, not the singular "shard" in PR 61).

I meant keep the "server" (that's the logical name of the connector), and add an extra key, maybe just "task_key" since you are passing it. If you want to add "shards", you might want to pass the actual shards?

shichao-an:
> What is the new content I need to put in for the previous 2 commit messages? Or should I squash all the 4 commits into one commits? From the console output, it seems it is complaining about the three extra files? Run ! git log --format='format:%s' refs/remotes/origin/$BASE_REF.. | tail -n +2 | grep -Ev '^(DBZ-[[:digit:]]+)|(\[release\])|(\[jenkins-jobs\]) ' Remove three extra files Merge branch 'DBZ-5382' of https://github.com/HenryCaiHaiying/debezium-connector-vitess into DBZ-5382 Error: Process completed with exit code 1. On Monday, August 8, 2022, 02:48:59 PM PDT, Hossein Torabi ***@***.***> wrote: Its seems the CI fails because of first 2 commit messages. Please use this doc to fix it: https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/changing-a-commit-message#:~:text=On%20the%20command%20line%2C%20navigate,message%2C%20and%20save%20the%20commit. — Reply to this email directly, view it on GitHub, or unsubscribe. You are receiving this because you authored the thread.Message ID: ***@***.***>

I think you need to add the DBZ-???? prefix to all commits

src/main/java/io/debezium/connector/vitess/VitessConnector.java
shichao-an: For 1, it seems that the blindly reading old offsets approach definitely has some flows, such as that it couldn't tell whether the old offset was  the exactly the "previous generation". If you 

Sounds like specifying previous tasks.max will do. Can you elaborate how do you make this work? 
- Users have to remember the previous tasks.max all the time right?
- If they didn't specify previous tasks.max what would happen?

Another question I have is:
Does tasks.max=N guarantees N tasks will be created? From our previous discussions in the JIRA ticket I remember it was not the case, e.g. when number of shard, M < tasks.max, then only M tasks will be created, right? Will this become a blocker for future migration?

HenryCaiHaiying: For approach 1 to check previous partition keys in offset storage, the ideal way is to convince Kafka Connect people to add a iterator API in OffsetStorageReader, but that might take a couple of month to add that API and trickle back to this project.  In the interim, as a workaround we might just have to guess the previous partition key names by looping through task0_1, task0_2, task0_3 ..., this might get expensive for large number of partitions;

For approach 2: asking user to pass in previous task numbers.  If we want to task, we want to ask the previous number of tasks used (instead of previous tasks.max specified since the connector might create less number of tasks), the user can get the number of tasks used by querying on connector REST API before migration.

There is also approach 3: asking user to specify a generation number.  The partition key will take the form of taskId_generationNum.  if task keys for the current generation (e.g. 2) is not present (i.e. task0_2, task1_2, task2_2 ... not present) but we found task keys for the previous generation (i.e. task0_1) present, then this is a an error;

Or approach 4: ask user to specify a boolean argument: offsetExisted.  On connector startup, if we found no partitions for the current taskId_totalTasks but offsetExisted=true, it's an error;  Or reversely, if we found partition keys existed but offSetExisted=false it's also an error.   On first run of the connector, the user will specify tasks.max=initial_value, offsetExisted=false;  On the 2nd run, the user will need to update the param to: tasks.max=initial_value, offSetsExisted=true;   When the tasks.max is changing, the user will do the offset migration in kafka topic and update the task param to tasks.max=new_value, offSetsExisted=true;   If the user forgot to do the offset migration and start the connector with the above value, the connector will fail to start since it cannot find offset in the storage for that tasks.max.

src/main/java/io/debezium/connector/vitess/VitessPartition.java
HenryCaiHaiying: That 'server' name as the partition key is really useless.  Maybe the initial intention was to make sure there is no key conflicts when multiple databases are going into that Kafka offset topic.  But this partition key is listed under the connector name in the kafka topic.  When there are multiple databases or keyspaces, we would have to create multiple connectors (each with distinct name).  So inside kafka offset topic, it will become {connector1, {server='db1'}} {connector2, {server='db2'}}.  So under connector1, the dbServer is always going to be db1 (even when we start to create multiple source partitions to support multiple vitess shards), that's why I don't intend to keep dbServer name as part of the source partition name.  And that dbServer name is already persisted in another kafka config: connect-configs, so that information is not lost either.

HenryCaiHaiying:
> 

Let me try to add the DBZ-??? prefix to all commits

src/main/java/io/debezium/connector/vitess/VitessPartition.java
shichao-an: For the offset topic, you are right, it's kind of useless.

However, the second part of the offset key actually is called "source partitions", i.e {server='db1'} part in `{connector1, {server='db1'}}`. See https://github.com/debezium/debezium/blob/main/debezium-core/src/main/java/io/debezium/pipeline/spi/Partition.java

I think it's used by Debezium event dispatcher to decide the partition key for the source record. https://github.com/debezium/debezium/blob/main/debezium-core/src/main/java/io/debezium/pipeline/EventDispatcher.java#L397

That means source partition is better to be a unique one per connector and task to avoid unexpected behavior such as hot spot. it's better to keep the original semantics while incrementally adding more connector-specific semantics without breaking original ones. cc @gunnarmorling @jpechane

HenryCaiHaiying: This SourcePartition concept is actually not a debezium thing, it's a concept in Kafka Connect Framework: https://github.com/apache/kafka/blob/trunk/connect/api/src/main/java/org/apache/kafka/connect/source/SourceRecord.java#L33.   Kafka Connect use a Map of keys to identify a Source Partition to allow for situations that partition key is a combo key (e.g. {database=db1, table=table1}).  In the vitess connector, if we use the offset.storage.per.task mode, the partition key is based on the taskId, the database name doesn't really mean much for the source partition, the database name is not a secondary key used to further drill into the partition space.  The Map for the source offset is designed for the similar flexibility purpose.  The keys in the map for source offset doesn't need to be the same as the keys in the Map for the source partition.

HenryCaiHaiying: To both debezium and Kafka Connect, the SourcePartition and SourceOffset are pass-through items, it's only meaningful within the vitess connector.  As long as their use are consistent between writing and retrieving, the content of source partition is totally decided by vitess connector.  I think I can change the key name within SourcePartition to be 'taskId' to reflect the current partitioning scheme, database name is not really used in this partitioning scheme to identify anything so it doesn't really need to be included.

shichao-an: I see what you meant. I don't have a strong opinion. 

My thought is that even if source partition a is passed through entity or for debug/information purpose, I still see its value if you have multiple connectors in the same Kafka Connect cluster with the same multi-task distribution, say 4 tasks, it will let the downstream components to believe that they come from the same "source partition", but in fact they are not; they come from multiple different connectors. In this case I think "make every field meaningful" makes more sense to me.

HenryCaiHaiying: OK, let me add a task_key param in SourcePartitions.

src/main/java/io/debezium/connector/vitess/VitessConnector.java
HenryCaiHaiying: @shichao-an To move this issue forward, I think I will probably just proceed with approach 3: (a) has the partition key to take the form of <taskId>_<numTasks>_<genNum>, (b) and the user to specify the generation number for the current partitioning, genNum starts with 1 and will increase continuously for each new partitioning change (genNumber=0 is reserved for the current partitioning which only uses server key) (c) for the user who wants to handle offset automatic migration from previous task partitioning, she will specify previous numTasks param, so we should be able to find all the existing partition keys in offset storage using <taskId>_<prevNumTasks>_<genNum-1>;  if those previous partitioning keys doesn't exist, the program will error out;   (d) for the user who doesn't want do offset auto migration (e.g. some of the old keys doesn't exist anymore), she can specify offset_migration=false, so the program will basically take the current offset (the tail of the binlog queue) from Vitess ;  Although user needs to specify a partitioning generation number and a previousNumTasks, this approach is more precise during migration and it solves the offset auto migration (an obstacle I tried to avoid previously).

shichao-an: The generation number approach sounds more robust to me. I wanted to bring it out since you mentioned it earlier, but somehow I forgot :) I will defer the implementation details to you.

src/main/java/io/debezium/connector/vitess/VitessPartition.java
HenryCaiHaiying: @shichao-an Please take a look at today's commit: https://github.com/debezium/debezium-connector-vitess/pull/91/commits/8c3975e0ef8c012784c510f6058a6f8337ba702f which addressed this issue (adding task_key param) plus more unit/integration tests.

HenryCaiHaiying:
@shichao-an I made the commit 3 which adds a generation number into the partition key and handles the offset auto migration when the task parallelism changes.  I think commit 2 and 3 addresses all the review comments, please take a look.  There are many files changes from the 3 commits which might makes the review harder.  I am not sure whether you want me to merge in as 3 separate PRs or just one big PR.

HenryCaiHaiying:
Can someone review this PR to move this forward? @shichao-an @jpechane @Naros

src/main/java/io/debezium/connector/vitess/VitessPartition.java
```diff
@@ -10,20 +10,31 @@
...
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import io.debezium.pipeline.spi.Partition;
 import io.debezium.util.Collect;
 
 public class VitessPartition implements Partition {
-    private static final String SERVER_PARTITION_KEY = "server";
+    private static final Logger LOGGER = LoggerFactory.getLogger(VitessPartition.class);
```
shichao-an, line 20: nit: I think logger is not used in this class?

src/main/java/io/debezium/connector/vitess/VitessPartition.java
```diff
@@ -58,7 +69,16 @@ static class Provider implements Partition.Provider<VitessPartition> {
 
         @Override
         public Set<VitessPartition> getPartitions() {
-            return Collections.singleton(new VitessPartition(connectorConfig.getLogicalName()));
+            if (connectorConfig.offsetStoragePerTask()) {
+                String taskKey = connectorConfig.getVitessTaskKey();
+                if (taskKey == null) {
+                    throw new RuntimeException("No vitess.task.key in: {}" + connectorConfig.getConfig());
```
shichao-an, line 75:
```suggestion
                    throw new RuntimeException("No vitess.task.key in: " + connectorConfig.getConfig());
```

src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
```diff
@@ -170,11 +227,32 @@ public class VitessConnectorConfig extends RelationalDatabaseConnectorConfig {
...
+    // We copy the definition here instead of importing the class from connect.runtime package
+    protected static final String TASKS_MAX_CONFIG = "tasks.max";
+
+    // The vitess.task.key config, the value is in the form of <task-id>_<num-tasks>,
+    // VitessConnector will populate the value of this param and pass on to VitessConnectorTask
+    protected static final String VITESS_TASK_KEY_CONFIG = "vitess.task.key";
+
+    // The vitess.task.shards config, the value is a comma separated vitess shard names
+    // VitessConnector will populate the value of this param and pass on to VitessConnectorTask
+    protected static final String VITESS_TASK_KEY_SHARDS_CONFIG = "vitess.task.shards";
```
shichao-an, line 248: VITESS_TASK_SHARDS_CONFIG?

src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
```diff
@@ -170,11 +227,32 @@ public class VitessConnectorConfig extends RelationalDatabaseConnectorConfig {
...
+
+    // The vgtid assigned to the given task in the json format, this is the same format as we would see
+    // in the Kafka offset storage.
+    // e.g. [{\"keyspace\":\"ks\",\"shard\":\"-80\",\"gtid\":\"MySQL56/0001:1-114\"},
+    // {\"keyspace\":\"ks\",\"shard\":\"80-\",\"gtid\":\"MySQL56/0002:1-122\"}]
+    protected static final String VITESS_KEY_KEY_VGTID_CONFIG = "vitess.task.vgtid";
```
shichao-an, line 254: VITESS_TASK_VGTID_CONFIG?

src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
```diff
@@ -274,4 +352,28 @@ public boolean includeUnknownDatatypes() {
...
+
+    public List<String> getVitessTaskKeyShards() {
+        return getConfig().getStrings(VITESS_TASK_KEY_SHARDS_CONFIG, ",");
+    }
+
+    public Vgtid getVitessTaskVgtid() {
+        String vgtidStr = getConfig().getString(VITESS_KEY_KEY_VGTID_CONFIG);
+        return vgtidStr == null ? null : Vgtid.of(vgtidStr);
+    }
+}
```
shichao-an, line 379: nit: newline

src/main/java/io/debezium/connector/vitess/VitessConnector.java
```diff
@@ -20,31 +26,143 @@
...
+                LOGGER.info("No previous offset for partition: {}, fall back to only server key", par);
+                par = new VitessPartition(connectorConfig.getLogicalName(), null);
+                offset = offsetStorageReader.offset(par.getSourcePartition());
+            }
+            if (offset == null) {
+                if (expectsOffset) {
+                    throw new IllegalArgumentException(String.format("No offset found for %s", par));
+                }
+                else {
+                    LOGGER.warn("No offset found for key key: {}", taskKey);
```
shichao-an, line 76:
```suggestion
                    LOGGER.warn("No offset found for task key: {}", taskKey);
```

src/main/java/io/debezium/connector/vitess/VitessConnector.java
```diff
@@ -20,31 +26,143 @@
...
+            final int gen = connectorConfig.getOffsetStorageTaskKeyGen();
+            Map<String, String> prevGtidsPerShard = gen > 0 ? getGtidPerShardFromStorage(prevNumTasks, gen - 1, true) : null;
+            LOGGER.info("Previous gtids Per shard: {}", prevGtidsPerShard);
+            int tasks = Math.min(maxTasks, shards.size());
+            LOGGER.info("There are {} vitess shards for maxTasks: {}, we will use {} tasks",
+                    shards.size(), maxTasks, tasks);
+            if (gen > 0 && tasks == prevNumTasks) {
+                throw new IllegalArgumentException(String.format(
+                        "Previous num.tasks: %s and current num.tasks: %s are the same",
+                        prevNumTasks, tasks));
```
shichao-an, line 110: Maybe you should check this first at line 102 before calling `getGtidPerShardFromStorage`?

src/main/java/io/debezium/connector/vitess/VitessConnector.java
```diff
@@ -20,31 +26,143 @@
...
+            final int prevNumTasks = connectorConfig.getPrevNumTasks();
+            final int gen = connectorConfig.getOffsetStorageTaskKeyGen();
+            Map<String, String> prevGtidsPerShard = gen > 0 ? getGtidPerShardFromStorage(prevNumTasks, gen - 1, true) : null;
+            LOGGER.info("Previous gtids Per shard: {}", prevGtidsPerShard);
+            int tasks = Math.min(maxTasks, shards.size());
+            LOGGER.info("There are {} vitess shards for maxTasks: {}, we will use {} tasks",
+                    shards.size(), maxTasks, tasks);
+            if (gen > 0 && tasks == prevNumTasks) {
+                throw new IllegalArgumentException(String.format(
+                        "Previous num.tasks: %s and current num.tasks: %s are the same",
```
shichao-an, line 109: Maybe consider adding an explanation and hint for the next step such as "the previous generation must have different number of tasks. Choose a different max.tasks or new shards and retry"?

src/main/java/io/debezium/connector/vitess/VitessConnector.java
```diff
@@ -20,31 +26,143 @@
...
+                        prevNumTasks, tasks));
+            }
+            if (prevGtidsPerShard != null && prevGtidsPerShard.size() != shards.size()) {
+                throw new IllegalArgumentException(String.format("Different number of shards between %s and %S",
```
shichao-an, line 113: Same, maybe add an explanation and potential next step to fix this?

shichao-an (commented):
Sorry for the belated review.
Overall looks good. Left some comments

src/main/java/io/debezium/connector/vitess/VitessPartition.java
HenryCaiHaiying: Nice catch, fixed in the new commit.

src/main/java/io/debezium/connector/vitess/VitessPartition.java
HenryCaiHaiying: Nice catch, fixed in the new commit.

src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
HenryCaiHaiying: Nice catch, fixed in the new commit.

src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
HenryCaiHaiying: Nice catch, fixed in the new commit.

src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
HenryCaiHaiying: Nice catch, fixed in the new commit.

src/main/java/io/debezium/connector/vitess/VitessConnector.java
HenryCaiHaiying: Nice catch, fixed in the new commit.

src/main/java/io/debezium/connector/vitess/VitessConnector.java
HenryCaiHaiying: Nice catch, fixed in the new commit.

src/main/java/io/debezium/connector/vitess/VitessConnector.java
HenryCaiHaiying: Nice catch, fixed in the new commit.

src/main/java/io/debezium/connector/vitess/VitessConnector.java
HenryCaiHaiying: Nice catch, fixed in the new commit.

HenryCaiHaiying:
@shichao-an Thanks for your comments on 8/23.  I just pushed out another commit which took in all your suggestions, please take a look.

shichao-an (approved):
LGTM. Thank you for your work!

shichao-an:
Once merged, you can update the latest documentation with the added features and configs:
https://github.com/debezium/debezium/blob/main/documentation/modules/ROOT/pages/connectors/vitess.adoc

HenryCaiHaiying:
@shichao-an Thanks for the review.  I will update the doc once the PR is merged.

HenryCaiHaiying:
@shichao-an I just made another commit because we just discovered a regression when PUT REST call is made to update configs of an existing connector, should be a simple review:

    DBZ-5382: Fix VitessConnector.validateAllFields() when called from Rest API

shichao-an (approved):

HenryCaiHaiying:
@jpechane Since shichao approved the PR, will you be able to merge in this PR?

jpechane:
@HenryCaiHaiying Applied, thanks a lot! Could you please provide a docs update PR to the main repo too? Thanks

HenryCaiHaiying:
See corresponding doc PR: https://github.com/debezium/debezium/pull/3830

PR #96, opened 2022-09-01
DBZ-5572: Specify table.include.list during VStream subscription

HenryCaiHaiying:
When table.include.list was used to only subscribe to changes to a subset of tables in the database, the filtering is currently done in debezium VM.  This is less efficient comparing to the filtering at the VtTablet level.   During VStream subscription, you can specify which tables you are interested, the VtTable will only send the changes to these tables to you.  The benefit of doing filtering in VtTable level is:

1. There are much less network bytes sending over the network,
2. it also has the advantage of avoiding errors related to tables you are not interested.  For example, we had seen errors related to schema mismatch on gh-ost (online schema migration) metadata table, we are not really interested in those tables.  By specifying to subscribe to only the data tables we are interested in, we can avoid those problems as well.

The fix is relatively straightforward by using Binlogdata.Filter during VStreamRequest building.  The details of those filters can be found in: https://github.com/vitessio/vitess/blob/release-14.0/go/vt/vttablet/tabletserver/vstreamer/planbuilder.go#L316 .   Note that when you provide the table name for the filtering, you also need to provide a SELECT query to further restrict the rows/columns from that table, we are currently not filtering on row/column level.

New integration tests are added to cover the table.include.list filtering.

The change only applies when you specify table.include.list config params.

For reference, sometimes we saw the errors not related to our tables like below:

Errors like the following (that _ghc table is the temp table created during gh-ost migration process)
VStream streaming onError. Status: Status{code=UNKNOWN, description=target: byuser.-4000.replica: vttablet: rpc error: code = Unknown desc = stream (at source tablet) error @ 08fb1cf3-0ce5-11ed-b921-0a8939501751:1-1443715: unknown table _mentions_unread_ghc in schema, cause=null}

### DIFF
--- a/src/main/java/io/debezium/connector/vitess/VitessConnector.java
+++ b/src/main/java/io/debezium/connector/vitess/VitessConnector.java
@@ -12,6 +12,8 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 
 import org.apache.kafka.common.config.ConfigDef;
 import org.apache.kafka.common.config.ConfigValue;
@@ -25,6 +27,7 @@ import io.debezium.config.Configuration;
 import io.debezium.connector.common.RelationalBaseSourceConnector;
 import io.debezium.connector.vitess.connection.VitessReplicationConnection;
 import io.debezium.relational.RelationalDatabaseConnectorConfig;
+import io.debezium.util.Strings;
 import io.grpc.StatusRuntimeException;
 import io.vitess.proto.Query;
 import io.vitess.proto.Vtgate;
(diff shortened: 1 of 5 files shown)


HenryCaiHaiying:
@jpechane @shichao-an One more fix for vitess connector, please take a look.

jpechane:
@HenryCaiHaiying Thanks for the PR, theis is a neat idea. The issue is that the table list is not list of table names but list f *regular exepressions* so the correct implementation should obtain the list of tables and apply the regex list (using `Predicates` class` to obtain the list of tables.

HenryCaiHaiying:
@jpechane I pushed the 2nd command which addressed your concern.  table.include.list is now processed as a list of patterns instead of direct table names.

jpechane:
@HenryCaiHaiying Good job! Applied

PR #108, opened 2022-11-23
DBZ-4720 Validate Debezium Server configuration properties

ggaborg:


jpechane:
@ggaborg Applied, thanks!

PR #112, opened 2022-12-13
DBZ-5930: Support snapshot feature without an automatic retry

yoheimuta:
ref. https://issues.redhat.com/browse/DBZ-5930

This PR adjusts the event state machine in gRPC responses to support the snapshot feature.

The following two cases differ from the streaming and should be taken care of.

- Duplicate VGTID when the VStream Copy starts
- Duplicate BEGIN when the VStream Copy starts and the table has no rows

### DIFF
--- a/src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
+++ b/src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
@@ -96,7 +96,9 @@ public class VitessConnectorConfig extends RelationalDatabaseConnectorConfig {
             .withImportance(ConfigDef.Importance.HIGH)
             .withDescription(
                     "Single GTID from where to start reading from for a given shard."
-                            + " It has to be set together with vitess.shard");
+                            + " It has to be set together with vitess.shard."
+                            + " If set an empty string, the connector starts copying the tables for the given shard first."
+                            + " If not configured, the connector streams changes from the latest position for the given shard.");
 
     public static final Field TABLET_TYPE = Field.create(VITESS_CONFIG_GROUP_PREFIX + "tablet.type")
             .withDisplayName("Tablet type to get data-changes")
--- a/src/main/java/io/debezium/connector/vitess/connection/VitessReplicationConnection.java
+++ b/src/main/java/io/debezium/connector/vitess/connection/VitessReplicationConnection.java
@@ -116,9 +116,16 @@ public class VitessReplicationConnection implements ReplicationConnection {
                         case VGTID:
                             // We always use the latest VGTID if any.
                             if (newVgtid != null) {
-                                LOGGER.warn("Received more than one VGTID events and the previous one is {}. Using the latest: {}",
-                                        newVgtid.toString(),
-                                        event.getVgtid().toString());
+                                if (newVgtid.getRawVgtid().getShardGtidsList().stream().findFirst().map(s -> s.getTablePKsCount() == 0).orElse(false)
+                                    && event.getVgtid().getShardGtidsList().stream().findFirst().map(s -> 0 < s.getTablePKsCount()).orElse(false)) {
+                                    LOGGER.info("Received more than one VGTID events during a copy operation and the previous one is {}. Using the latest: {}",
+                                            newVgtid.toString(),
+                                            event.getVgtid().toString());
+                                } else {
+                                    LOGGER.warn("Received more than one VGTID events and the previous one is {}. Using the latest: {}",
+                                            newVgtid.toString(),
+                                            event.getVgtid().toString());
+                                }
                             }
                             newVgtid = Vgtid.of(event.getVgtid());
                             break;
@@ -131,8 +138,18 @@ public class VitessReplicationConnection implements ReplicationConnection {
                             }
                             if (beginEventSeen) {
                                 String msg = "Received duplicate BEGIN events";
-                                setError(msg);
-                                return;
+                                // During a copy operation, we receive the duplicate event once when no record is copied.
+                                String eventTypes = bufferedEvents.stream().map(VEvent::getType).map(Objects::toString).collect(Collectors.joining(", "));
+                                if (eventTypes.equals("BEGIN, FIELD, VGTID")) {
+                                    msg += String.format(" during a copy operation. No harm to skip the buffered events. Buffered event types: %s",
+                                            eventTypes);
+                                    LOGGER.info(msg);
+                                    reset();
+                                }
+                                else {
+                                    setError(msg);
+                                    return;
+                                }
                             }
                             beginEventSeen = true;
                             break;
--- a/src/test/java/io/debezium/connector/vitess/VitessConnectorIT.java
+++ b/src/test/java/io/debezium/connector/vitess/VitessConnectorIT.java
@@ -734,6 +734,43 @@ public class VitessConnectorIT extends AbstractVitessConnectorTest {
         assertEquals(new HashSet<>(expectedTables), tables);
     }
 
+    @Test
+    public void testCopyAndReplicateTable() throws Exception {
+        TestHelper.executeDDL("vitess_create_tables.ddl");
+        TestHelper.execute(INSERT_NUMERIC_TYPES_STMT, TEST_UNSHARDED_KEYSPACE);
+        String tableInclude = TEST_UNSHARDED_KEYSPACE + "." + "numeric_table";
+        startConnector(builder -> builder.with(VitessConnectorConfig.GTID, ""),
+                false, false, 1, -1, -1, tableInclude);
+
+        // We should receive a record written before starting the connector.
+        int expectedRecordsCount = 1;
+        consumer = testConsumer(expectedRecordsCount);
+        consumer.await(TestHelper.waitTimeForRecords(), TimeUnit.SECONDS);
+        SourceRecord record = assertRecordInserted(topicNameFromInsertStmt(INSERT_NUMERIC_TYPES_STMT), TestHelper.PK_FIELD);
+        assertSourceInfo(record, TEST_SERVER, TEST_UNSHARDED_KEYSPACE, "numeric_table");
+        assertRecordSchemaAndValues(schemasAndValuesForNumericTypes(), record, Envelope.FieldName.AFTER);
+
+        // We should receive additional record from numeric_table
+        consumer.expects(expectedRecordsCount);
+        assertInsert(INSERT_NUMERIC_TYPES_STMT, schemasAndValuesForNumericTypes(), TestHelper.PK_FIELD);
+    }
+
+    @Test
+    public void testCopyNoRecordsAndReplicateTable() throws Exception {
+        TestHelper.executeDDL("vitess_create_tables.ddl");
+
+        String tableInclude = TEST_UNSHARDED_KEYSPACE + "." + "numeric_table";
+        // An exception due to duplicate BEGIN events shouldn't occur
+        startConnector(builder -> builder.with(VitessConnectorConfig.GTID, ""),
+                false, false, 1, -1, -1, tableInclude);
+
+        int expectedRecordsCount = 1;
+        consumer = testConsumer(expectedRecordsCount);
+
+        // We should receive record from numeric_table
+        assertInsert(INSERT_NUMERIC_TYPES_STMT, schemasAndValuesForNumericTypes(), TestHelper.PK_FIELD);
+    }
+
     private void testOffsetStorage(boolean offsetStoragePerTask) throws Exception {
         TestHelper.executeDDL("vitess_create_tables.ddl", TEST_UNSHARDED_KEYSPACE);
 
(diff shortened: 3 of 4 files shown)


src/main/java/io/debezium/connector/vitess/connection/VitessReplicationConnection.java
```diff
@@ -116,9 +116,16 @@ public void onNext(Vtgate.VStreamResponse response) {
                         case VGTID:
                             // We always use the latest VGTID if any.
                             if (newVgtid != null) {
-                                LOGGER.warn("Received more than one VGTID events and the previous one is {}. Using the latest: {}",
-                                        newVgtid.toString(),
-                                        event.getVgtid().toString());
+                                if (newVgtid.getRawVgtid().getShardGtidsList().stream().findFirst().map(s -> s.getTablePKsCount() == 0).orElse(false)
+                                    && event.getVgtid().getShardGtidsList().stream().findFirst().map(s -> 0 < s.getTablePKsCount()).orElse(false)) {
+                                    LOGGER.info("Received more than one VGTID events during a copy operation and the previous one is {}. Using the latest: {}",
```
yoheimuta, line 121: The actual log is:

```
2022-12-12 18:57:05,028 INFO   ||  Received more than one VGTID events during a copy operation and the previous one is [{"keyspace":"test_unsharded_keyspace","shard":"0","gtid":"MySQL56/4edb14ad-79c3-11ed-a748-0242ac110002:1-947"}]. Using the latest: shard_gtids {
  keyspace: "test_unsharded_keyspace"
  shard: "0"
  gtid: "MySQL56/4edb14ad-79c3-11ed-a748-0242ac110002:1-947"
  table_p_ks {
    table_name: "t1"
    lastpk {
      rows {
        lengths: 1
        values: "1"
      }
    }
  }
}
```

jpechane:
@yoheimuta Hi, thanks for the PR. It looks relatively easy to be done. For the sake of cross-connector consistency, could you please take a look at `snapshot.mode` config option used by other connectors? If possible I'd like to drive/configure snapshotting via it rather than with a special value of GTID setting.

yoheimuta:
Thank you for telling me!
I'm looking into `snapshot.mode` and `formatting.`

src/test/java/io/debezium/connector/vitess/VitessConnectorIT.java
```diff
@@ -820,15 +855,24 @@ private void startConnector(boolean hasMultipleShards) throws InterruptedExcepti
...
                                 int numTasks, int gen, int prevNumTasks, String tableInclude)
             throws InterruptedException {
+        startConnector(customConfig, hasMultipleShards, offsetStoragePerTask, numTasks, gen, prevNumTasks,
+                tableInclude, VitessConnectorConfig.SnapshotMode.NEVER);
+    }
+
+    private void startConnector(Function<Configuration.Builder, Configuration.Builder> customConfig,
+                                boolean hasMultipleShards, boolean offsetStoragePerTask,
+                                int numTasks, int gen, int prevNumTasks, String tableInclude,
+                                VitessConnectorConfig.SnapshotMode snapshotMode)
```
yoheimuta, line 872: [memo] When you give `null` as snapshotMode here, the connector uses the default snapshotMode (INITIAL).

yoheimuta:
@jpechane Hi, I fixed these two points.

- formating: [87a8c12](https://github.com/debezium/debezium-connector-vitess/pull/112/commits/87a8c12a371ea60237c74f1692fcd3b13d31a0b5)
- snapshot.mode: [7cfb6a0](https://github.com/debezium/debezium-connector-vitess/pull/112/commits/7cfb6a0c814a163c2e1d948bb400010a2af72424) [f2c59a5](https://github.com/debezium/debezium-connector-vitess/pull/112/commits/f2c59a56b264a929d3b6bfbc3a9166ff58643079) [b12e5b7](https://github.com/debezium/debezium-connector-vitess/pull/112/commits/b12e5b78278b7058dc698eaf63c9f8f632389893)

Could you please review this PR?

src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
```diff
@@ -38,6 +39,71 @@ public class VitessConnectorConfig extends RelationalDatabaseConnectorConfig {
     private static final String VITESS_CONFIG_GROUP_PREFIX = "vitess.";
     private static final int DEFAULT_VTGATE_PORT = 15_991;
 
+    /**
+     * The set of predefined SnapshotMode options or aliases.
+     */
+    public static enum SnapshotMode implements EnumeratedValue {
```
yoheimuta, line 45: I took a look at https://debezium.io/documentation/reference/2.0/connectors/mongodb.html#mongodb-property-snapshot-mode.

src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
```diff
@@ -210,6 +278,17 @@ public class VitessConnectorConfig extends RelationalDatabaseConnectorConfig {
                             + "Once we persist the new offsets in offset storage using new partition key "
                             + "based on current <numTasks> and <gen>, we will no longer read prev.num.tasks param");
 
+    public static final Field SNAPSHOT_MODE = Field.create("snapshot.mode")
+            .withDisplayName("Snapshot mode")
+            .withEnum(SnapshotMode.class, SnapshotMode.INITIAL)
```
yoheimuta, line 283: All the other connectors set the default to INITIAL.

src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
```diff
@@ -296,6 +376,9 @@ public String getShard() {
     }
 
     public String getGtid() {
+        if (getSnapshotMode() == SnapshotMode.INITIAL) {
```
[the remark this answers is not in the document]
HenryCaiHaiying: Also the caller of getGtid is VitessReplicationConnection.defaultVGtid().  There are three code paths in defaultVGtid() (offsetStoragePerTask mode, shard = "", shard is not empty)  config.getGtid() only covers one of the paths, can the gtid="" also works for the other two paths?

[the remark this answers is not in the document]
jpechane: @HenryCaiHaiying I don't think this is wrong. Other connectors also execute full table scans upon connector start as they do the snapshot of existing content.

jpechane:
@yoheimuta One key question, could you please add a test that will verify what will happen after connector restart? I am a bit afraid that if the state of snapshot being completed is not recorded in offsets then it will be reexcuted again but I might be wrong.
Still it is is important to test it.

src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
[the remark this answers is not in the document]
yoheimuta: > can the gtid="" also works for the other two paths?
>> shard = ""

This case is not supported by the current Vitess now. I reported it to the Vitess team. 
The fix will be included in Vitess 16 by https://github.com/vitessio/vitess/issues/11886.

So, I added a warning output and a test for now [70a89be](https://github.com/debezium/debezium-connector-vitess/pull/112/commits/70a89bece3f17f27e7ceb394687f602903d76e9a).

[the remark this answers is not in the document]
yoheimuta: >> offsetStoragePerTask mode

Thanks! offsetStoragePerTask mode slipped my mind.
Maybe it should be supported. I'm looking into it.

yoheimuta:
@jpechane Thanks! It makes sense.
I'm adding a test that verifies what will happen after the connector restart.

yoheimuta:
@jpechane Hi, I added the test: [689a2c5](https://github.com/debezium/debezium-connector-vitess/pull/112/commits/689a2c59724664faa403c91929970eec6ae6f8f7).

src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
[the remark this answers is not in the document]
HenryCaiHaiying: I have some concerns on the defaulting behavior.  The defaulting snapshot pull will change the existing system's behavior.  The current default is to pull from the tail of the BinLog, now when people upgrade to the newer version of debezium and if they are not careful, it becomes pull from the head of the BinLog with a big performance impact.

The other issue is the GTID in the config is being ignored when the snapshot mode is Initial (which is the default).  Sometimes we use that GTID config parameter to have the connector to rewind to an old position in the binlog, now this function is lost if we forgot to change the snapshot mode.  I think the system should honor the GTID param if it's set, only use "" if gtd is not present (flip your current fallback logic)

I kind of like the first version of the PR where the user will explicitly set the GTID param to "" to indicate he wants to pull from the beginning of the binlog.  This wouldn't cause confusions or performance impact if not careful.

src/main/java/io/debezium/connector/vitess/connection/VitessReplicationConnection.java
```diff
@@ -352,6 +370,9 @@ public static Vgtid defaultVgtid(VitessConnectorConfig config) {
         }
         else {
             if (config.getShard() == null || config.getShard().isEmpty()) {
+                if (config.getGtid().equals("")) {
```
HenryCaiHaiying, line 373: Although vitess doesn't support empty shard and empty gtid position, there is a workaround possible in debezium connector side.  We can get all the shards from vtgate and replace the empty shard param with all shards and continue (this was actually what was happening in VtGate code when you don't specify shard).  VitessConnector.getVitessShards() can get you all the shards.

src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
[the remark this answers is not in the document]
yoheimuta: > offsetStoragePerTask mode
>> Maybe it should be supported. I'm looking into it.

I added a commit [822a58d](https://github.com/debezium/debezium-connector-vitess/pull/112/commits/822a58dc357842d425b8e73336b174fbe5377d55) to support it.

src/main/java/io/debezium/connector/vitess/connection/VitessReplicationConnection.java
yoheimuta: That sounds nice!
I was slanted toward that the connector side code was simpler and we would wait for the fix in Vitess.

But, given that [this use case is recommended at the document](https://debezium.io/documentation/reference/2.0/connectors/vitess.html#vitess-property-shard) and the current code has `VitessConnector.getVitessShards(),` I changed my mind that it's worth implementing the workaround now.
So, I'm getting to it.

src/main/java/io/debezium/connector/vitess/VitessConnector.java
```diff
@@ -150,7 +150,7 @@ public List<Map<String, String>> taskConfigs(int maxTasks, List<String> shards)
                     }
                     if (gtidStr == null) {
                         LOGGER.warn("No previous gtid found either for shard: {}, fallback to current", shard);
```
HenryCaiHaiying, line 152: Can you change the order of line 152 and 153, so you can print out the gtidStr value in the LOGGER.warn?

yoheimuta: Thanks! I overlooked it.
Added [5ea8780](https://github.com/debezium/debezium-connector-vitess/pull/112/commits/5ea878007cb93f43d5bf8ec74bb0193def13eff0).

HenryCaiHaiying: Thanks, looks good to me.

HenryCaiHaiying:
@yoheimuta The change looks good to me overall, thanks for adding support for empty shard and OffsetStoragePerTask mode.  I have some concerns on changing the initial pull behavior to snapshot the whole table instead of from tail because of the performance impact , but if you guys feel this is consistent with the other connectors, you can move forward but the release note has to clearly specify this changing behavior and the potential performance impact on the database side.

jpechane:
@HenryCaiHaiying According to the test the initial snapshot should not be executed after connector upgrade. In that case the current installations are safe and the new installations will have the same behaviour as the other connectors.
Yes, it will be mentioned in release notes under breaking changes section.

yoheimuta:
@HenryCaiHaiying @jpechane Thank you for your careful review!
I agree with clearly specifying this breaking change behavior.

@jpechane So, should I provide a docs update PR to the main repo as the next step?

HenryCaiHaiying:
@yoheimuta Thanks for the contribution.  It was nice someone can push though the changes on both debezium and vitess community.

jpechane:
@yoheimuta Applied, thanks! yes, please prepare the docs PR too.

yoheimuta:
See corresponding doc PR: https://github.com/debezium/debezium/pull/4158

PR #115, opened 2023-01-11
Test of CI

jpechane:


ani-sha:
A quick change is needed here: https://github.com/debezium/debezium-connector-vitess/pull/116#discussion_r1066907062

PR #119, opened 2023-01-15
DBZ-6011 Vitess: Handle the shard list difference between current db shards and persisted shards

HenryCaiHaiying:
When dbzium connector went offline for some time and Vitess shard splits during that time (e.g. s1 split into s10 and s11), the current shard list from v$session will have latest (s10, s11) however the shard list persisted Kafka offset topic only contains old shard s1 (pointed at an old position gt1).

When vitess connector restarts, the current logic in VitessConnector.taskConfigs() will use the latest shard list (s10, s11) to do the task assignment and it will use "current" (i.e. tail of the binlog queue) for s10/s11.  The correct behavior should use the old shard (s1) and old position (e.g. gt1) from persisted Kafka offset storage, this way the connector would subscribe to the exact point when it was stopped before.

When the vtgate continue playing the binlog events from vttablet, it will eventually encounter the shard split binlog event where tablet stream from s1 vttablet will be closed and be replaced with tablet streams from s10 and s11. All these will happen transparently for vitess connector, vitess connector will seamlessly receives the new events from new shards from vtgate (as if it was connected online all the time).

The fix is to enhance the logic iin VitessConnector.taskConfigs() to favor the shards from persisted offset storage.  However there are a few caveats:

1. If the tasks persisted in offset storage is empty (this can happen the first time when the task assignment initiated for the current generation of task number), we fall back to the shards from the previous generation of task assignment.

2. When the task persited in offset storage for the current gen is not complete (this can happen when some task worker crashed before it even gets the chance to persist its offset), we would need to fall back to use the current db shard list from v$session.

3. In the case the shards from offset storage diverges from the ones from current db list from v$session (they don't form a superset relationship), we would have to use the ones from offset storage.  But in this case we would need to require all tasks are persisted in offset storage to make sure we didn't miss any shards, otherwise the connector had to abort and wait for manual intervention.

Several new unit tests are added to handle the new edge cases.

### DIFF
--- a/src/main/java/io/debezium/connector/vitess/VitessConnector.java
+++ b/src/main/java/io/debezium/connector/vitess/VitessConnector.java
@@ -97,14 +97,14 @@ public class VitessConnector extends RelationalBaseSourceConnector {
         return taskConfigs(maxTasks, shards);
     }
 
-    public List<Map<String, String>> taskConfigs(int maxTasks, List<String> shards) {
-        LOGGER.info("Calculating taskConfigs for {} tasks and shards: {}", maxTasks, shards);
+    public List<Map<String, String>> taskConfigs(int maxTasks, List<String> currentShards) {
+        LOGGER.info("Calculating taskConfigs for {} tasks and shards: {}", maxTasks, currentShards);
         if (connectorConfig.offsetStoragePerTask()) {
             final int prevNumTasks = connectorConfig.getPrevNumTasks();
             final int gen = connectorConfig.getOffsetStorageTaskKeyGen();
-            int tasks = Math.min(maxTasks, shards.size());
+            int tasks = Math.min(maxTasks, currentShards.size());
             LOGGER.info("There are {} vitess shards for maxTasks: {}, we will use {} tasks",
-                    shards.size(), maxTasks, tasks);
+                    currentShards.size(), maxTasks, tasks);
             if (gen > 0 && tasks == prevNumTasks) {
                 throw new IllegalArgumentException(String.format(
                         "Previous num.tasks: %s and current num.tasks: %s are the same. "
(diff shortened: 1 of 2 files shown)


src/main/java/io/debezium/connector/vitess/VitessConnector.java
```diff
@@ -113,18 +113,53 @@ public List<Map<String, String>> taskConfigs(int maxTasks, List<String> shards)
...
-                        prevGtidsPerShard, shards));
+            if (prevGtidsPerShard != null && prevGtidsPerShard.size() != currentShards.size()) {
+                LOGGER.warn("Some shards for the previous generation {} are not persisted.  Expected shards: {}",
+                        prevGtidsPerShard.keySet(), currentShards);
             }
             final String keyspace = connectorConfig.getKeyspace();
+            // Check the task offsets for the current gen, the offset might not be persisted if this gen just turned on
             Map<String, String> gtidsPerShard = getGtidPerShardFromStorage(tasks, gen, false);
-            if (gtidsPerShard != null && gtidsPerShard.size() != shards.size()) {
+            if (gtidsPerShard != null && gtidsPerShard.size() != currentShards.size()) {
```
jpechane, line 126: @HenryCaiHaiying The code lGTM.One question though. Is it possible that while Debezium is down that there will be split and later on merge so the shards are different but the count is the same?

HenryCaiHaiying: In theory this can happen but in practice, the chances are very low due to:
1. debezium down is not likely going over multiple hours otherwise this realtime change capture pipeline downstream users will be in trouble;  And if debezium down for too long, it will have problems on reconnecting back to Vitess DB since the gtid position it stored previously might already be removed from Vitess side due to Vitess data retention policy;
2. Vitess shard split is a somewhat likely event but merge back is a very rare event.  In our company (Slack) of using Vitess for more than 3 years, there is probably only once or twice merge back event.
3. Vitess shard split (or merge back) is usually a planned event executed manually.  The chance of vitess shards split and then merge back during the debezium down time without notice is very unlikely.

HenryCaiHaiying: But if you feel strongly we should check against this situation, I can also change the comparison code to do that.

jpechane: Even if it is rare, can't it lead into a hard-to-identify issue? If yes then I'd prefer to hedge against it too.

HenryCaiHaiying: OK, I've added a new method hasSameShards() to do element-wise content comparison (instead of just relying on the size check), please take a look @jpechane

HenryCaiHaiying: Ping @jpechane ...

jpechane:
@HenryCaiHaiying Applied, thanks a lot!

PR #120, opened 2023-01-19
DBZ-5852:Snapshotter#snapshotCompleted is invoked regardless of snapshot result

adasari:


jpechane:
@adasari Applied, thanks a lot!

PR #121, opened 2023-01-26
DBZ-6043: Vitess: Support Mapping unsigned bigint mysql column type to long

HenryCaiHaiying:
The columns in MySQL/Vitess is defined as `unsigned bigint(20)`, currently Vitess Connector is mapping this column type as JDBC.String type and it becomes `string` in Avro schema when publishing to schema registry.  However, the debezium changelog output is integrated with Hudi for bootstrap/delta stream integration.  Hudi is mapping this column type as `long` on the bootstrapping path which caused the schema mismatch.  We would prefer to support mapping `unsigned bigint` to `long` in vitess debezium connector as a configuration option.

In theory, unsigned bigint has the range of  0 to 18,467,440,737,095,551,615 and long has the range of  -9,223,372,036,854,775,808 to 9,223,372,036,854,775,807 .  However `long` type would be OK in practice since those columns usually stores the ids and ids don't go over to 9,223,372,036,854,775,807.  Many data systems (e.g. Hudi) chose to do a rough mapping since it's difficult to find an unsigned bigint type in either Java or Avro.

This problem was also encountered in Vitess MySQL Connector and over there a config option `bigint.unsigned.handling.mode` was introduced to support mapping bigint.unsigned to long: https://debezium.io/documentation/reference/stable/connectors/mysql.html#mysql-property-bigint-unsigned-handling-mode

We are following the example of MySQL Connection for Vitess and introduced the same config option: bigint.unsigned.handling.mode

The option takes in the following two choice values:
* string (the default): the bigint unsigned will be mapped to string
* long: the bigint unsigned will be mapped to long

VitessType.resolveType is enhanced to handle the mapping of UINT64 (the Vitess VStream representation of unsigned bigint) to either JDBC.BigInt (becomes long later) or JDBC.Varchar (becomes string later)

Unit tests are added for this.


jpechane:
@HenryCaiHaiying Could you please add also support for `BigDecimal`? In that way this will be superset of MySQL config options. And I'd say it will be better from consistency point of view.

HenryCaiHaiying:
@jpechane  Per your review comment, added `Precise` mode to make it compatible/superset of MySQLConverter's BigIntUnsigned mode.  Now there are three modes to choose from:
    * String (default): treat the column as string/varchar type
    * Long: treat the column as long/BigInt type
    * Precise: treat the column as BigDecimal
    
Move the handling up into VitessValueConverter (very similar to the ones in MySQLValueConverter)

src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
```diff
@@ -289,6 +355,16 @@ public static SnapshotMode parse(String value, String defaultValue) {
...
 
+    public static final Field BIGINT_UNSIGNED_HANDLING_MODE = Field.create("bigint.unsigned.handling.mode")
+            .withDisplayName("BIGINT UNSIGNED Handling")
+            .withEnum(BigIntUnsignedHandlingMode.class, BigIntUnsignedHandlingMode.STRING)
+            .withGroup(Field.createGroupEntry(Field.Group.CONNECTOR, 1))
+            .withWidth(Width.SHORT)
+            .withImportance(ConfigDef.Importance.MEDIUM)
+            .withDescription("Specify how BIGINT UNSIGNED columns should be represented in change events, including: "
+                    + "'string' (the default) represent values using Java's 'string'; "
+                    + "'long' represents values using Java's 'long', which may not offer the precision but will be far easier to use in consumers.");
```
jpechane, line 366: One option is missing

jpechane (approved):
@HenryCaiHaiying  LGTM, thanks! I left a small comment related to the description in the config option. Could you please also prepare a PR with docs update?

src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
HenryCaiHaiying: Made another PR to add the description for 'precise' mode for the option description

HenryCaiHaiying:
Added Doc PR: https://github.com/debezium/debezium/pull/4242

jpechane:
@HenryCaiHaiying Applied, thanks!

HenryCaiHaiying:
@jpechane Can this PR be integrated into 2.1 branch?

jpechane:
@HenryCaiHaiying Hi, I am sorry. This is not a bug but new feature and we strictly limit new features backported to stable branches as they are aimed to be ideally bugix only. We are willing to provide exceptions in super important cases ut this one does not look like it.

HenryCaiHaiying:
In this case, when will you create 2.2 branch?

jpechane:
@HenryCaiHaiying Current main is the 2.2. The separate 2.2 will be created with the final release. Main repo has 2.2 only to generate documentation.

PR #134, opened 2023-04-19
DBZ-6356 Support streaming a list of shards and accompanying GTIDs

twthorn:
### Summary
With the Vitess connector, the shard setting can only stream a single shard or all shards. It would be helpful to extend this to support a shard list such that gradual scale up/roll out can be done safely with the Vitess connector. This is read in as a CSV list and the accompanying GTIDs can also be specified.

### Verification
Added unit tests & an integration test verifying we can read multiple shards when they are specified as CSV config.

Note: We implement this feature specifically for the single task mode currently. In a subsequent PR we will implement support for streaming a list of shards and handling multiple tasks. Want to keep the scope of this one just to single task and multi shard streaming.

### DIFF
--- a/src/main/java/io/debezium/connector/vitess/VitessConnector.java
+++ b/src/main/java/io/debezium/connector/vitess/VitessConnector.java
@@ -220,9 +220,10 @@ public class VitessConnector extends RelationalBaseSourceConnector {
             if (maxTasks > 1) {
                 throw new IllegalArgumentException("Only a single connector task may be started");
             }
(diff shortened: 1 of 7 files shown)


HenryCaiHaiying:
> This is read in as a CSV list 

Just to clarify, this is a comma-separated config string, not a CSV file input.  It would also be good to provide a config example.

jpechane:
@twthorn Applied, thanks a lot! I just introduced few constants to make the code a bit readable.

PR #135, opened 2023-04-20
DBZ-6365 Support streaming a list of shards/gtids with multiple tasks

twthorn:
### Summary

We added support previously for single thread mode for reading in a string of csv shards. Now we extend that to also work with multiple tasks

### Verification
Added unit tests & acceptance tests for verifying things work as expected when multiple tasks are used with the shard config csv string.
Note: for the registration of the metrics in the integration test, they were always registered under `task_0_1_0` despite the fact that `task_0_2_0` would be the expected name for a connector with 2 max tasks. When I checked the logs the `taskConfigs` function was always being called with 1 even if our config said 2. From what I can tell this is a limitation of the integration tests using the embedded engine, so I just overrode the numTasks for generating the task ID to always be 1.

### DIFF
--- a/src/main/java/io/debezium/connector/vitess/VitessConnector.java
+++ b/src/main/java/io/debezium/connector/vitess/VitessConnector.java
@@ -17,6 +17,7 @@ import java.util.Objects;
 import java.util.Set;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 
 import org.apache.kafka.common.config.ConfigDef;
 import org.apache.kafka.common.config.ConfigValue;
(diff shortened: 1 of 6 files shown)


src/main/java/io/debezium/connector/vitess/VitessConnector.java
```diff
@@ -142,17 +143,22 @@ public List<Map<String, String>> taskConfigs(int maxTasks, List<String> currentS
                         prevGtidsPerShard.keySet(), currentShards);
             }
             final String keyspace = connectorConfig.getKeyspace();
+            // Check the configs in case there is a user specified GTID override
+            verifyShardGtidConfig();
+            Map<String, String> gtidsPerShard = getGtidsPerShardFromConfig();
```
HenryCaiHaiying, line 148: This logic is not correct, the shard/grids will need to come from the Kafka storage instead of from static config on the normal run.

The static config value is not used for the first run when there is no history in Kafka storage.

HenryCaiHaiying: Please don't merge in this PR yet, I will work with Tom separately to sort out some of the issues.

src/main/java/io/debezium/connector/vitess/VitessConnector.java
```diff
@@ -220,17 +226,32 @@ else if (gtidsPerShard != null && !currentShards.containsAll(gtidsPerShard.keySe
...
+                    .boxed()
+                    .collect(Collectors.toMap(connectorConfig.getShard()::get, connectorConfig.getGtid()::get));
+            LOGGER.info("Found GTIDs per shard in config {}", gtidsPerShard);
+        }
+        return gtidsPerShard;
+    }
+
+    private void verifyShardGtidConfig() {
+        final List<String> gtids = connectorConfig.getGtid();
+        if (connectorConfig.getShard() != null &&
```
jpechane, line 247: This is re-used check - maybe it would be good to extract it it to a separate method at `VitessConnectorConfig`

twthorn: Added, thanks!

src/main/java/io/debezium/connector/vitess/VitessConnector.java
twthorn: Updated with this feedback, thanks!

twthorn:
Updated to use GTIDs only on initial run (not stored in previous or current shard-gtid map). Also use shard list whenever specified. Added more tests. Also force-pushed to fix commit message formatting (no other change).

src/main/java/io/debezium/connector/vitess/VitessConnector.java
```diff
@@ -123,23 +131,30 @@ public List<Map<String, String>> taskConfigs(int maxTasks, List<String> currentS
...
-                        prevNumTasks, tasks));
+                        prevNumTasks, tasks, previousShards, currentShards));
             }
-            // Check the task offsets persisted from previous gen, we expect the offsets are saved
-            Map<String, String> prevGtidsPerShard = gen > 0 ? getGtidPerShardFromStorage(prevNumTasks, gen - 1, true) : null;
-            LOGGER.info("Previous gtids Per shard: {}", prevGtidsPerShard);
             if (prevGtidsPerShard != null && !hasSameShards(prevGtidsPerShard.keySet(), currentShards)) {
                 LOGGER.warn("Some shards for the previous generation {} are not persisted.  Expected shards: {}",
                         prevGtidsPerShard.keySet(), currentShards);
+                if (prevGtidsPerShard.keySet().containsAll(currentShards)) {
```
HenryCaiHaiying, line 153: Should we throw Exception in this case?   People usually don't look at the logs until problem happens.    If we let this pass, we won't have the old gtids for shrinked shards anymore since we don't look for older gtids more than one generation old.   If we think shrinking shards is a valid use case, then we probably need to a config flag to indicate whether we should halt on shard list shrinking or not, the default value for that config value should be false.

HenryCaiHaiying (commented):
Looks good overall, one comment on whether we should support shrinking shard list.

src/main/java/io/debezium/connector/vitess/VitessConnector.java
HenryCaiHaiying: Thanks for the update, it looks good to me now.

HenryCaiHaiying (approved):

src/main/java/io/debezium/connector/vitess/VitessConnector.java
twthorn: Thanks for the feedback. Updated to throw an exception.

If we end up with a use case later, we can implement functionality for contracting the shard list, but for now we will opt for the more cautious path of preventing any lost state (plus this is a new feature, so there's no use cases that do this contracting shard list, and we only know of a use case for expansion i.e., ours).

twthorn:
Fixed a bug where the shards were not being passed in from the config correctly between the two taskConfigs methods. Updated tests to reproduce the error, which is now fixed.

Not sure why commit message is failing, I have the correct prefix.

HenryCaiHaiying:
@twthorn Not sure if you title the commit as 'DBZ-6365: ' instead of 'DBZ-6365 ' would make the commit check pass.

jpechane:
@twthorn Applied, thanks!

PR #138, opened 2023-05-30
DBZ-6510 Use event processing failure handling mode in VitessReplicationConnection

twthorn:
### Summary

Use the event processing failure handling mode for handling errors appropriately in the VitessReplicationConnection. This is similar to what is done for [Postgres](https://github.com/debezium/debezium/blob/main/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/PostgresReplicationConnection.java#L457). We follow the standard behaviors for error/skip/warn/ignore.

### Verification

Add a few integration tests for each branch of error/skip/warn/ignore. Check that the error throwable is set/not set appropriately. Use the example use case of an oversized grpc message to be the guaranteed failure (by setting byte max very low).

### DIFF
--- a/src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
+++ b/src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
@@ -175,15 +175,6 @@ public class VitessConnectorConfig extends RelationalDatabaseConnectorConfig {
         }
     }
 
-    public EventProcessingFailureHandlingMode getEventProcessingFailureHandlingMode() {
-        if (super.getEventProcessingFailureHandlingMode() == null) {
-            return EventProcessingFailureHandlingMode.FAIL;
-        }
-        else {
-            return super.getEventProcessingFailureHandlingMode();
-        }
-    }
-
     public static final Field VTGATE_HOST = Field.create(DATABASE_CONFIG_PREFIX + JdbcConfiguration.HOSTNAME)
             .withDisplayName("Vitess database hostname")
             .withType(Type.STRING)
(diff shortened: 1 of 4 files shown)


src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
```diff
@@ -175,6 +175,15 @@ public static BigIntUnsignedHandlingMode parse(String value, String defaultValue
         }
     }
 
+    public EventProcessingFailureHandlingMode getEventProcessingFailureHandlingMode() {
```
jpechane, line 178: This should be unnecessary as `io.debezium.config.CommonConnectorConfig.EVENT_PROCESSING_FAILURE_HANDLING_MODE` declares `FAIL` as the default value.

src/test/java/io/debezium/connector/vitess/VitessReplicationConnectionIT.java
```diff
@@ -48,6 +48,172 @@ public void beforeEach() {
...
+        connection.startStreaming(
+                startingVgtid,
+                (message, vgtid, isLastRowEventOfTransaction) -> {
+                    if (!started.get()) {
+                        started.set(true);
+                    }
+                    consumedMessages.add(new MessageAndVgtid(message, vgtid));
+                },
+                error);
+        // Since we are using the "current" as the starting position, there is a race here
```
jpechane, line 84: This is a bit unsafe. Could you please take a look at `io.debezium.connector.mysql.StreamingSourceIT.inconsistentSchema(EventProcessingFailureHandlingMode)` to see how to intercept log messages and wait for them?

jpechane (changes requested):
@twthorn Thanks a lot. I left two comments related to removal of unnecessary code and change in tests. The functionality itself LGTM.

twthorn:
@jpechane Thank you for the review. I've updated with the feedback to refactor the test to wait on log messages, and to remove the unnecessary config method.

jpechane:
@twthorn Applied, thanks!

PR #141, opened 2023-06-28
DBZ-6617 Add shard to source

twthorn:
Add shard to source which is present in row events. Allows downstream users to determine what shard this event is for (we already include keyspace/table so we also include the shard). More context in [DBZ-6617](https://issues.redhat.com/browse/DBZ-6617).

### DIFF
--- a/src/main/java/io/debezium/connector/vitess/connection/VStreamOutputMessageDecoder.java
+++ b/src/main/java/io/debezium/connector/vitess/connection/VStreamOutputMessageDecoder.java
@@ -148,6 +148,7 @@ public class VStreamOutputMessageDecoder implements MessageDecoder {
         else {
             String schemaName = schemaTableTuple[0];
             String tableName = schemaTableTuple[1];
+            String shard = rowEvent.getShard();
             int numOfRowChanges = rowEvent.getRowChangesCount();
             int numOfRowChangesEventSeen = 0;
             for (int i = 0; i < numOfRowChanges; i++) {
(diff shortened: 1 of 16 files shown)


src/main/java/io/debezium/connector/vitess/connection/VStreamOutputMessageDecoder.java
```diff
@@ -148,21 +148,22 @@ private void decodeRows(Binlogdata.VEvent vEvent, ReplicationMessageProcessor pr
         else {
             String schemaName = schemaTableTuple[0];
             String tableName = schemaTableTuple[1];
+            String shard = rowEvent.getShard();
```
mfvitale, line 151: How are you managing the possible 'UnsupportedOperationException`? What do you think to return an empty string in case no shard?

mfvitale:
@twthorn Thanks for contributing. Generally LGTM but I have only one concern about `UnsupportedOperationException`.

src/main/java/io/debezium/connector/vitess/connection/VStreamOutputMessageDecoder.java
twthorn: This is calling `getShard` on the vitess-grpc-client `RowEvent` so it won't return the `UnsupportedOperationException` since that's a custom defined exception in Debezium. [RowEvent will return an empty string](https://github.com/vitessio/vitess/blob/main/go/vt/proto/binlogdata/binlogdata.pb.go#L1404-L1409) by default.


For cases where we might raise the `UnsupportedOperationException`, this happens when we access an implementation of `ReplicationMessage`, which is only ever done [here](https://github.com/debezium/debezium-connector-vitess/pull/141/files#diff-9bb425f66b65357858125fd2a8f03290fc3efcd6dcb9a355e7fa3219823fa018R117). That line is only reached if it's [not a `TransactionalMessage`](https://github.com/debezium/debezium-connector-vitess/pull/141/files#diff-9bb425f66b65357858125fd2a8f03290fc3efcd6dcb9a355e7fa3219823fa018R94), and [neither a `DDLMessage` nor an `OtherMessage`](https://github.com/debezium/debezium-connector-vitess/pull/141/files#diff-9bb425f66b65357858125fd2a8f03290fc3efcd6dcb9a355e7fa3219823fa018R107). Those are the only three classes that would throw the exception, so we know that when `getShard` is called it will be safe (i.e., done for a [VStreamOutputReplicationMessage](https://github.com/debezium/debezium-connector-vitess/pull/141/files#diff-93d3bda8235cccc7f5592262a96a3ed1504e7dec63335ab39b20a648a222d94f))

mfvitale:
@twthorn Applied, thanks!

twthorn:
@mfvitale Thank you for the help!

One thing to note: This should probably be included in the debezium release notes since it adds a new required field (could have some impact on downstream consumers who expect the previous schema). It looks like Jira tickets for changes like this one should have the label `add-to-upgrade-guide`. I don't have permission to do this or to comment this context, so I'm mentioning this here.

mfvitale:
@twthorn  Can you please provide documentation change in another PR? Thanks.

PR #145, opened 2023-07-19
DBZ-6680 Add setting to disable transaction topic

twthorn:
This is an initial version. There are a few parts we need to address mentioned in https://issues.redhat.com/browse/DBZ-6680

This handles disabling the transaction topic. However, we may want to do this in debezium-core, if we want to support this for all connectors. ie also a  check a common setting here https://github.com/debezium/debezium/blob/main/debezium-core/src/main/java/io/debezium/pipeline/txmetadata/TransactionMonitor.java#L135


jpechane:
Thanks for the PR, I think we should discuss the final desing before moving forward with it.

PR #148, opened 2023-08-02
DBZ-6748 Make VARCHAR binary collation field types have a VitessType of VARCHAR

twthorn:
For VARCHAR columns with a binary collate function (ends in `_bin`) the SourceRecord output type is `BYTES`. In order to stay consistent with the [documentation](https://debezium.io/documentation/reference/stable/connectors/vitess.html#vitess-basic-types) and how [MySQL](https://debezium.io/documentation/reference/stable/connectors/vitess.html#vitess-basic-types) connector handles this, we need to ensure that we have an output type of `STRING`. Add the additional logic needed to do this. There are two cases when QUERY.FIELD type is a VARBINARY
1. The column type may be varchar (this is a varchar column with binary collation) in which case return varchar type
2. Column type is not varchar (e.g., a typical varbinary column), then fall through to next case and return varbinary as usual

### DIFF
--- a/src/main/java/io/debezium/connector/vitess/VitessType.java
+++ b/src/main/java/io/debezium/connector/vitess/VitessType.java
@@ -101,9 +101,6 @@ public class VitessType {
                 if (field.getColumnType().toUpperCase().contains("VARCHAR")) {
                     return new VitessType(type, Types.VARCHAR);
                 }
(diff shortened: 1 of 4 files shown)


jpechane:
@twthorn Applied, thanks a lot!

PR #149, opened 2023-08-07
DBZ-6775 Apply schema changes per shard

twthorn:
Schema changes happen per shard, so we should update the schemas accordingly in order to allow for us to correctly decode messages received for each shard without errors. First, add tests to reproduce the issues seen in [DBZ-6775](https://issues.redhat.com/browse/DBZ-6775)/[Zulip](https://debezium.zulipchat.com/#narrow/stream/348255-community-vitess/topic/Vitess.20Connector.20Error), then add the logic needed to resolve this. We add shard as a catalog, and use table string mapper to ensure our include/exclude list doesn't use this catalog/shard value, and we only use it for storing schemas.

Note: I looked into a way around not using the catalog variable as the shard. It seems quite complex to avoid this. There were a couple approaches I evaluated that ended up not being feasible:
1. Instead only store table IDs with a keyspace & table name, and then have a wrapper class that maps shards to `VitessDatabaseSchemas`
 - This means we aren't storing the shard as the catalog. However, when we first create the [dispatcher](https://github.com/debezium/debezium-connector-vitess/blob/main/src/main/java/io/debezium/connector/vitess/VitessConnectorTask.java#L95) we need to give it the database schema it will use for filtering/retrieving schemas. Therefore the table ID that is sent on the [dispatch call](https://github.com/debezium/debezium/blob/main/debezium-core/src/main/java/io/debezium/pipeline/EventDispatcher.java#L252) must have the shard information represented in it. So we cannot avoid using catalog as shard in this approach (or some other method e.g, concatenating shard to keyspace)
2. Make a custom data collection ID implementation (e..g, `VitessTableId` that has shard info represented)
- This is the idea based on the above learning that we need shard to be represented by the table ID in order for the dispatcher to use it correctly. So we could implement the interface for [DataCollectionId](https://github.com/debezium/debezium/blob/main/debezium-api/src/main/java/io/debezium/spi/schema/DataCollectionId.java) and have a properly constructed Id with fields for shard/keyspace/table. The issue here that we use several other classes which rely on us using the standard TableId and TableImpl (e.g., RelationalDatabaseSchema, Tables.java, etc.). There is lots of shared code from debezium-core that we use because we are a standard table/relational connector. We would need to reimplement all that.

### DIFF
--- a/src/main/java/io/debezium/connector/vitess/Filters.java
+++ b/src/main/java/io/debezium/connector/vitess/Filters.java
@@ -28,8 +28,8 @@ public class Filters {
         // Define the filter using the include/exclude list for table names ...
         this.tableFilter = Tables.TableFilter.fromPredicate(
                 Selectors.tableSelector()
-                        .includeTables(config.tableIncludeList())
-                        .excludeTables(config.tableExcludeList())
+                        .includeTables(config.tableIncludeList(), new VitessTableIdToStringMapper())
+                        .excludeTables(config.tableExcludeList(), new VitessTableIdToStringMapper())
                         .excludeSchemas(SYSTEM_SCHEMA_EXCLUDE_LIST)
                         .build());
     }
--- a/src/main/java/io/debezium/connector/vitess/VitessDatabaseSchema.java
+++ b/src/main/java/io/debezium/connector/vitess/VitessDatabaseSchema.java
@@ -84,4 +84,18 @@ public class VitessDatabaseSchema extends RelationalDatabaseSchema {
     public static TableId parse(String table) {
         return TableId.parse(table, false);
     }
+
+    /**
+     * Create a TableId where we use the shard name as the catalog name.
+     * This is needed to ensure that schema updates are applied to each shard individually.
+     * This is used in conjunction with {@link VitessTableIdToStringMapper} which excludes the shard/catalog name when
+     * determining if a table is in the include/exclude list.
+     * @param shard The shard of the table ID to build
+     * @param keyspace The keyspace of the table ID to build
+     * @param table The table name of the tbale ID to build
+     * @return TableId with shard (as catalog), keyspace (as schema), and table
+     */
+    public static TableId buildTableId(String shard, String keyspace, String table) {
+        return new TableId(shard, keyspace, table);
+    }
 }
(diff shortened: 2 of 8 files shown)


jpechane:
@twthorn Applied, thanks a lot. I think it is good engineering decision. I was thinking about `VitessTableId` to extend `TableId` but at the end it would not probably add any added value and in fact catalog is apt mapping name for shard as each shard has its own view (catalog) of database schema.

twthorn:
@jpechane Thank you for the review. One thing to note: In the release notes it should be mentioned that this could change the output topic names.

If the connector is run with the default [topic.naming.strategy](https://debezium.io/documentation/reference/stable/connectors/vitess.html#vitess-property-topic-naming-strategy) `io.debezium.schema.SchemaTopicNamingStrategy` then it will have no effect (since it doesn't use catalog, only uses schema/table parts). However, if the connector is being run with a different topic naming strategy (e.g.,[io.debezium.schema.DefaultTopicNamingStrategy](https://github.com/debezium/debezium/blob/main/debezium-core/src/main/java/io/debezium/schema/DefaultTopicNamingStrategy.java#L33)) that includes catalog name in the topic name, then it would result in shards being added to topic names.

On a related note, being able to output data for specific Vitess shards to separate topics was actually a use case we were considering contributing back upstream. So, I think it's a valid use case, and it's an easy work around for users to pick another topic naming strategy if they don't want to use the shard in the topic name.

jpechane:
@twthorn Thanks a lot, would you be willing to send a PR to https://github.com/debezium/debezium.github.io/blob/develop/releases/2.4/release-notes.asciidoc ?

twthorn:
@jpechane Sure thing, here's the PR https://github.com/debezium/debezium.github.io/pull/935

I have also created a new ticket for easing the migration of users who were using `DefaultTopicNamingStrategy` [DBZ-6800](https://issues.redhat.com/browse/DBZ-6800) to add a new topic naming strategy that will exclude the vitess shard. I put out a PR https://github.com/debezium/debezium-connector-vitess/pull/151 to add this new topic naming strategy vitess connector users can use to migrate.

I thought about adding the topic naming strategy in the main Debezium repo, but this seems like it would only be applicable to Vitess connectors, where we just want to drop shard information and only include table name (the behavior from before when catalog was null and not the shard). So I added it in the Vitess connector repo, but if we want it elsewhere I can add it there.

PR #151, opened 2023-08-10
DBZ-6800 Add TableTopicNamingStrategy

twthorn:
See [DBZ-6800](https://issues.redhat.com/browse/DBZ-6800) for more details

Add a topic naming strategy that can remove the shard information for users of the vitess connector. Also maintain parity with previous versions such that users do not need to change their topic.prefix and lose connector state.


jpechane:
@twthorn Good idea, thanks!

PR #152, opened 2023-08-21
DBZ-6830 Switch INFO level to DEBUG for partial & multi-response transactions

twthorn:
See https://issues.redhat.com/browse/DBZ-6830 these logs can be very frequent for large keyspaces with many multi-response/partial transactions and clutter the logs. Similar to other logic in this class, if a log message could be logged for each message received, it should be a debug level log not info.


jpechane:
@twthorn Applied, thank

PR #154, opened 2023-08-31
DBZ-6852 Always fail/reset GRPC channel on error

twthorn:
It is not possible to skip messages if they are over the [max inbound message size](https://grpc.github.io/grpc-java/javadoc/io/grpc/ManagedChannelBuilder.html#maxInboundMessageSize(int)) on the channel. If an error is received, the channel doesn't continue streaming, so we should always error out and reset.

### DIFF
--- a/src/main/java/io/debezium/connector/vitess/connection/VitessReplicationConnection.java
+++ b/src/main/java/io/debezium/connector/vitess/connection/VitessReplicationConnection.java
@@ -223,7 +223,7 @@ public class VitessReplicationConnection implements ReplicationConnection {
 
             @Override
             public void onError(Throwable t) {
-                LOGGER.info("VStream streaming onError. Status: " + Status.fromThrowable(t), t);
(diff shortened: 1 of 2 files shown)


src/main/java/io/debezium/connector/vitess/connection/VitessReplicationConnection.java
```diff
@@ -223,31 +223,10 @@ public void onNext(Vtgate.VStreamResponse response) {
...
-                            break;
-                    }
-                }
-                else {
-                    LOGGER.error("VStream streaming onError. Status: {}", status, t);
-                    // Only propagate the first error
-                    error.compareAndSet(null, t);
-                    reset();
-                }
+                LOGGER.info("VStream streaming onError. Status: " + Status.fromThrowable(t), t);
```
jpechane, line 226: IMHO this should be `error` not `info` level.

jpechane:
@twthorn Applied, thanks

PR #158, opened 2023-09-20
DBZ-6944 Retry all StatusRuntimeException, use super method on others

twthorn:
We should retry on all StatusRuntimeException which are a catch all for all errors from the vstream grpc go client.

For throwables that are not that, we can use the super method (retries on IO exceptions only), same as other Debezium connectors (e.g., mysql).

### DIFF
--- a/src/main/java/io/debezium/connector/vitess/VitessErrorHandler.java
+++ b/src/main/java/io/debezium/connector/vitess/VitessErrorHandler.java
@@ -23,34 +23,9 @@ public class VitessErrorHandler extends ErrorHandler {
     protected boolean isRetriable(Throwable throwable) {
         if (throwable instanceof StatusRuntimeException) {
             final StatusRuntimeException exception = (StatusRuntimeException) throwable;
-            final String description = exception.getStatus().getDescription();
-            LOGGER.info("Exception code: {} and description: {}", exception.getStatus().getCode(), description);
-            switch (exception.getStatus().getCode()) {
-                case CANCELLED:
-                    // Try to match this description string:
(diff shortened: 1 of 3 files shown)


twthorn:
Some error on the VitessConnectorIT, will look into that

twthorn:
@jpechane All tests passing now, if you can give this a review. Thank you

src/main/java/io/debezium/connector/vitess/VitessErrorHandler.java
```diff
@@ -23,34 +23,9 @@ public VitessErrorHandler(VitessConnectorConfig connectorConfig, ChangeEventQueu
...
-                    // Stream timeout error due to idle VStream or vstream ended unexpectedly.
-                    if (description != null &&
-                            (description.equals("stream timeout") ||
-                                    description.contains("vstream ended unexpectedly") ||
-                                    description.contains("unexpected server EOF"))) {
-                        return true;
-                    }
-                    return false;
-            }
+            LOGGER.error("Exception status: {}", exception.getStatus(), exception);
```
jpechane, line 26: It should be `warn` not `error` as connector is recovering from it.

jpechane:
@twthorn LGTM, thanks! I left one comment related to log level, other than that I think we are good to go.

src/main/java/io/debezium/connector/vitess/VitessErrorHandler.java
jpechane: At the end I decided to keep it as is.

jpechane:
Applied, thanks a lot!

PR #163, opened 2023-10-17
DBZ-7050 Add automatic retry for snapshots

twthorn:
Add automatic retry for snapshots.

### Automatic Retry
As detailed in the [table copy RFC ](https://github.com/vitessio/vitess/issues/6277) the last PK value in the VGTID is used for resuming partially complete snapshots. We utilize this for our automatic retries and parse this additional information now being sent by Vitess in the latest version (i.e., including fixes like [this](https://github.com/vitessio/vitess/pull/11103)).

One implementation decision I debated between was switching all of our VGTID parsing logic to instead use the [provided protobuf json converters](https://protobuf.dev/reference/java/api-docs/com/google/protobuf/util/JsonFormat). This would simplify our codebase but also mean the Debezium logic is more tightly coupled with Vitesss/protobuf, so for the latter reason I decided against this.

### GTID -> VGTID config
Previous configs had the option of `vitess.gtid` which was originally added when `vitess.shard` only allowed for a single shard. With the changes to support multiple shards back in https://github.com/debezium/debezium-connector-vitess/pull/135, we made GTID into a CSV as well. In order to easily test the automatic retry snapshot behavior, it would be useful to also specify the last PK values in the config. Additionally, GTID csv adds unnecessary complexity to support another format of specifying GTID(s). So to resolve both of these issues, I changed this to simply be a `vitess.vgtid` config which is a JSON parseable string that we can use to initialize our VGTID for a VStream. Since VGTID is an array of keyspace/shard/gtid/lastpks it has far more versatility for the user to be able to arbitrarily specify the GTID(s) for any shard(s) subset and even PKs to resume snapshots on. This is what is used to test the snapshot resume behavior in the integration tests.

### DIFF
--- a/src/main/java/io/debezium/connector/vitess/TablePrimaryKeys.java
+++ b/src/main/java/io/debezium/connector/vitess/TablePrimaryKeys.java
@@ -1,19 +1,20 @@
 package io.debezium.connector.vitess;
 
-import binlogdata.Binlogdata;
-import com.fasterxml.jackson.annotation.*;
-
-import com.fasterxml.jackson.core.JsonProcessingException;
-import com.fasterxml.jackson.core.type.TypeReference;
-import com.fasterxml.jackson.databind.ObjectMapper;
-import com.google.protobuf.ByteString;
-import io.vitess.proto.Query;
-
 import java.io.UnsupportedEncodingException;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
 import java.util.stream.Collectors;
+
+import com.fasterxml.jackson.annotation.*;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.protobuf.ByteString;
+
+import io.vitess.proto.Query;
+
+import binlogdata.Binlogdata;
 
 public class TablePrimaryKeys {
 
@@ -35,7 +36,7 @@ public class TablePrimaryKeys {
     private final List<TableLastPrimaryKey> tableLastPrimaryKeys = new ArrayList<>();
 
     public TablePrimaryKeys(List<Binlogdata.TableLastPK> rawTableLastPrimaryKey) {
-        for (Binlogdata.TableLastPK tableLastPrimaryKey: rawTableLastPrimaryKey) {
+        for (Binlogdata.TableLastPK tableLastPrimaryKey : rawTableLastPrimaryKey) {
             this.rawTableLastPrimaryKeys.add(tableLastPrimaryKey);
             tableLastPrimaryKeys.add(new TableLastPrimaryKey(tableLastPrimaryKey.getTableName(), tableLastPrimaryKey.getLastpk()));
         }
@@ -43,14 +44,14 @@ public class TablePrimaryKeys {
 
     public static TablePrimaryKeys of(String tablePrimaryKeysJSON) {
         try {
-            List<TableLastPrimaryKey> tablePrimaryKeys = MAPPER.readValue(tablePrimaryKeysJSON, new TypeReference<List<TableLastPrimaryKey>>(){});
+            List<TableLastPrimaryKey> tablePrimaryKeys = MAPPER.readValue(tablePrimaryKeysJSON, new TypeReference<List<TableLastPrimaryKey>>() {
+            });
             return createFromTableLastPrimaryKeys(tablePrimaryKeys);
-        } catch (JsonProcessingException e) {
+        }
+        catch (JsonProcessingException e) {
             throw new IllegalStateException(e);
         }
     }
-
-
 
     public static TablePrimaryKeys createFromRawTableLastPrimaryKey(List<Binlogdata.TableLastPK> rawTableLastPrimaryKeys) {
         return new TablePrimaryKeys(rawTableLastPrimaryKeys);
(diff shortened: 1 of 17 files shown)


twthorn:
@jpechane Can you review this when you get the chance? Thank you!

src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
```diff
@@ -464,28 +467,34 @@ public List<String> getShard() {
         return getConfig().getStrings(SHARD, CSV_DELIMITER);
     }
 
-    public List<String> getGtid() {
+    public String getGtid() {
```
HenryCaiHaiying, line 470: Should the method name called `getVGtid`?

src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
```diff
@@ -222,22 +224,23 @@ public static BigIntUnsignedHandlingMode parse(String value, String defaultValue
...
-                            + "E.p. \"0\" for an unsharded keyspace. "
-                            + "Or \"-80\" for the -80 shard of the sharded keyspace.");
+                    "Shard(s) of which keyspace to read data from."
+                            + "E.g. \"0\" for an unsharded keyspace. "
+                            + "Or \"-80\" for the -80 shard of the sharded keyspace."
+                            + "Or \"-4000,4000-8000\" for two of the four shards of a sharded keyspace.");
 
-    public static final Field GTID = Field.create(VITESS_CONFIG_GROUP_PREFIX + "gtid")
-            .withDisplayName("gtid")
+    public static final Field VGTID = Field.create(VITESS_CONFIG_GROUP_PREFIX + "vgtid")
```
HenryCaiHaiying, line 232: Will it cause any regression with the config name changes from `gtid` to `vgtid`?

src/main/java/io/debezium/connector/vitess/TablePrimaryKeys.java
```diff
@@ -0,0 +1,361 @@
...
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.protobuf.ByteString;
+
+import io.vitess.proto.Query;
+
+import binlogdata.Binlogdata;
+
+public class TablePrimaryKeys {
```
jpechane, line 27: Could you please add JavaDoc to this class?

src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
```diff
@@ -37,8 +40,7 @@
  */
 public class VitessConnectorConfig extends RelationalDatabaseConnectorConfig {
 
-    public static final List<String> EMPTY_GTID_LIST = List.of(Vgtid.EMPTY_GTID);
-    public static final List<String> DEFAULT_GTID_LIST = List.of(Vgtid.CURRENT_GTID);
+    public static final String DEFAULT_GTID = Vgtid.CURRENT_GTID;
```
jpechane, line 43: Should this be moved to `Vgtid` class?

src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
jpechane: Yes, it will. It is necessary to deprecate the existing setting. So both should be available and when the deprecated one is used then WARN should be written to the log.

jpechane (commented):
@twthorn Thanks for the PR. I left few minor comments. One question for you. I supposed that incomplete snapshots would be resumed automaically so I'd expect additional data being writtent to or read from offsets. Is is the case?
Also for testing there should be test when connector is started. Then it is stopped in the middle of snapshot and started again. There should be no lost records or duplicates.

twthorn:
> I supposed that incomplete snapshots would be resumed automaically so I'd expect additional data being writtent to or read from offsets. Is is the case?

Correct, there will be additional data read/written from offsets. There is now an additional field `table_p_ks` that will always be read/written to offsets. When there is no table copy phase its value will be an empty list. When there is table copy phase, it includes the last sent primary keys of the table ([example](https://github.com/debezium/debezium-connector-vitess/pull/163/files#diff-e015639d76222b47af45d5ec1dd5069df5dd401c43ef34161b6f519885b57711R123-R141)). So it varies between marginally more data, and additional nested json array.

Also worth noting, for rolling forward, this works, and I added a test showing this (we can read offsets that lack the field `table_p_ks`). However, for rolling back, assuming offsets were written with newer code, the user may need to manually overwrite the offset data to remove the `table_p_ks` field to make it parseable. Jackson errors out on [unknown field](https://fasterxml.github.io/jackson-annotations/javadoc/2.6/com/fasterxml/jackson/annotation/JsonIgnoreProperties.html#ignoreUnknown()) by default. One option I considered was excluding the `table_p_ks` field when it's an empty list, but this means the vgtid format will be inconsistent/changing so I opted against it.

HenryCaiHaiying (approved):
lgtm

src/test/java/io/debezium/connector/vitess/VitessConnectorIT.java
```diff
@@ -836,6 +836,129 @@ public void testCopyAndReplicateTable() throws Exception {
...
+                assertThat(actualVgtid.getShardGtids().size()).isEqualTo(1);
+                assertThat(actualVgtid.getShardGtids().get(0).getTableLastPrimaryKeys()).isEqualTo(
+                        expectedVgtid.getShardGtids().get(0).getTableLastPrimaryKeys());
+            }
+        }
+    }
+
+    @Test
+    public void testSnapshotLargeTable() throws Exception {
+        TestHelper.executeDDL("vitess_create_tables.ddl");
```
jpechane, line 880: @twthorn Could you please store all the records captured and assert at the end that all are present and there are neither gaps nor duplicates?

jpechane:
@twthorn LGTM. I left one comment requesting a hardening of a a test. Otherwise we are good to go.

twthorn:
@jpechane Thanks for the changes! Please let me know if there's anything else I can do on the PR

jpechane:
@twthorn Applied, thanks a lot

PR #164, opened 2023-11-02
DBZ-7104 Fail the test always in the main thread

vjuranek:
Failing test in a different thread doesn't work reliably as JUnit may ignore exceptions (which is how failing test works) in other than main thread.

Do the assert and eventually fail the test in the main test thread.

https://issues.redhat.com/browse/DBZ-7104


jpechane:
@vjuranek Applied, thanks

PR #171, opened 2023-12-15
DBZ-7250 Set per-task vgtid within source task

twthorn:
If the GTID set grows large for a connector with many tasks, there will be timeouts since Kafka Connect imposes a non-configurable limit of [30 seconds](https://github.com/apache/kafka/blob/3.5.1/connect/runtime/src/main/java/org/apache/kafka/connect/storage/KafkaConfigBackingStore.java#L276) to write the configs of all tasks synchronously.

To avoid growing task configs, we do not add the vgtid offset information from storage until the task has received it's config (then we add it to maintain parity with existing logic and minimize refactoring).

This is a refactor that still uses the bulk of our current logic of reading offsets from storage/adding them to the task config. In subsequent work, it would be cleaner to refactor things so offsets are not in the config. We want offsets and configs to be decoupled, but right now they are coupled. Additionally, we should be able to automate the handling of generation number (similar to how other Debezium connectors do this). Opened [DBZ-7265](https://issues.redhat.com/browse/DBZ-7265) for this work

### Verification

I added unit tests & refactored existing tests to get all to pass.

Additionally, we had a failing connector in our internal deployment that was stuck with these timeout errors. Deploying these changes fixed it. Note: since we are changing the config format stored in the offset topic, it can cause Kafka Connect to fail. In this case users can delete the existing connector, and add a connector with the same name (will reuse the offsets, but purges the config history to get around Kafka connect's error). This should be noted as breaking change in the release notes.


twthorn:
The failure is not related to my changes, an error with apt-get/keys/signature of mysql repo. Maybe would work on retry but I don't have permissions to rerun. tests/itest passed locally so github workflow is having issues (I see recent upgrades so need to make sure those are all good).
```
[INFO] DOCKER> Err:5 http://repo.mysql.com/apt/debian buster InRelease
  The following signatures couldn't be verified because the public key is not available: NO_PUBKEY B7B3B788A8D3785C
[INFO] DOCKER> Reading package lists...
[INFO] DOCKER> 
[INFO] DOCKER> [91mW: GPG error: http://repo.mysql.com/apt/debian buster InRelease: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY B7B3B788A8D3785C
E: The repository 'http://repo.mysql.com/apt/debian buster InRelease' is not signed.

[INFO] DOCKER> Removing intermediate container 5b0156edfa6c
Error:  DOCKER> Unable to build image [vitess/test-database] : "The command '/bin/sh -c apt-get update' returned a non-zero code: 100"  ["The command '/bin/sh -c apt-get update' returned a non-zero code: 100" ]
```

HenryCaiHaiying (approved):
lgtm

jpechane:
Unfortunately given that we are doing the 2.5.0.Final this week and this is really breaking change I'd propose to move it to 2.6 to give it proper introduction period.
WRT build failure, Oracle's key used to sign MySQL packages is expired and a new one was not yet issued. Until that the build will stay broken as it seems there is no way how to force `apt` to ignor expired `GPG` keys.

twthorn:
@jpechane It looks like 2.5 was released, can we merge this now? Thanks!

jpechane:
Rebased on 2.6 to trigger CI. PR is going to be merged immediatlly after it is completed.
@twthorn Could you please add a comment to https://issues.redhat.com/browse/DBZ-7250 with text that I should add to release notes related to how the users are exepcted to upgrade? Thanks in advance!

jpechane:
@twthorn Applied, thanks!

PR #175, opened 2023-12-29
DBZ-7293 Fallback to old schema in case of race condition between ROW and FIELD

s-gelazevicius:
Sometimes due to race condition in Vitess, ROW and TYPE messages can be mixed up. Adding a fallback to the old schema version in case row length mismatches between Row and Table in schema cache.

https://vitess.io/docs/18.0/reference/vreplication/internal/tracker/#caveat

`Only best-effort versioning can be provided due to races between DDLs and DMLs`

### DIFF
--- a/src/main/java/io/debezium/connector/vitess/connection/VStreamOutputMessageDecoder.java
+++ b/src/main/java/io/debezium/connector/vitess/connection/VStreamOutputMessageDecoder.java
@@ -5,8 +5,6 @@
  */
 package io.debezium.connector.vitess.connection;
 
-import static io.debezium.connector.vitess.connection.ReplicationMessage.Column;
-
 import java.sql.Types;
 import java.time.Instant;
 import java.util.ArrayList;
@@ -47,8 +45,13 @@ public class VStreamOutputMessageDecoder implements MessageDecoder {
 
     private final VitessDatabaseSchema schema;
 
+    private final VitessDatabaseSchema schemaBackup;
+
     public VStreamOutputMessageDecoder(VitessDatabaseSchema schema) {
         this.schema = schema;
+        // Schema can be null. See: VitessConnector.validateConnection
+        if (schema != null) this.schemaBackup = schema.clone();
+        else this.schemaBackup = null;
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


src/main/java/io/debezium/connector/vitess/connection/VStreamOutputMessageDecoder.java
```diff
@@ -293,16 +296,32 @@ private Optional<Table> resolveRelation(String shard, String schemaName, String
...
+        return resolveColumns(row, table, false);
+    }
+
+    /**
+     *  Resolve the vEvent data to a list of replication message columns (with values).
+     *  NOTE: Sometimes due to race condition in Vitess ROW and TYPE messages can be mixed up, so there's a fallback
+     *  the old schema version in case row length mismatches between Row and Table in schema cache.
+     *  <a href="https://vitess.io/docs/18.0/reference/vreplication/internal/tracker/#caveat">...</a>
+     */
+    private List<Column> resolveColumns(Row row, Table table, Boolean fromBackup) {
```
jpechane, line 314:
```suggestion
    private List<Column> resolveColumns(Row row, Table table, boolean fromBackup) {
```

src/main/java/io/debezium/connector/vitess/connection/VStreamOutputMessageDecoder.java
```diff
@@ -47,8 +45,13 @@ public class VStreamOutputMessageDecoder implements MessageDecoder {
 
     private final VitessDatabaseSchema schema;
 
+    private final VitessDatabaseSchema schemaBackup;
+
     public VStreamOutputMessageDecoder(VitessDatabaseSchema schema) {
         this.schema = schema;
+        // Schema can be null. See: VitessConnector.validateConnection
+        if (schema != null) this.schemaBackup = schema.clone();
```
jpechane, line 53: There should be no cloning, both `schema` and `delayedSchema` should be passed from the constructor so it is apparent outside that two instances are needed.

src/main/java/io/debezium/connector/vitess/connection/VStreamOutputMessageDecoder.java
```diff
@@ -47,8 +45,13 @@ public class VStreamOutputMessageDecoder implements MessageDecoder {
 
     private final VitessDatabaseSchema schema;
 
+    private final VitessDatabaseSchema schemaBackup;
```
jpechane, line 48:
```suggestion
    private final VitessDatabaseSchema delayedSchema;
```

jpechane (commented):
Thanks for the PR. It looks good as a start.

Please make sure the code is properly formatted (run the `mvn install` locally).
I recommend to change the naming - instead of backup it should be delayed.
Also I belive the schema instance should be passed so it is visible that there migh be discrepancy in the schema processing.

Also please make sure the followinf scenarios are tested

* field added
* field removed
* field type changed

s-gelazevicius:
This doesn't fix race condition fully, it's still not handled in `io.debezium.pipeline.EventDispatcher`
Closing pull request

PR #179, opened 2024-01-29
DBZ-6993 Fix VitessConnectorTaskTest failure

shybovycha:
While working on https://github.com/debezium/debezium/pull/4973 I have noticed the unrelated test failures, one of them being `VitessConnectorTaskTest#shouldReadOffsetsWhenTaskOffsetStorageDisabled`.

The failure was pointing to logs not including the correct error message. Upon closer logs inspection, I have noticed a _very similar_ message is being logged, but the test included an extra message component. I removed it to make the tests green again.

N.B.: I am not very familiar with this codebase, so feel free to discard this PR.


jpechane:
@shybovycha Good catch, applied, thanks!

PR #183, opened 2024-03-10
DBZ-7618 Implement Versioned interfaces in Transformation and Converter plugins

harveyyue:
https://issues.redhat.com/browse/DBZ-7618


harveyyue:
no need this pr

PR #184, opened 2024-03-11
DBZ-7628 Use COMMIT time for events, not BEGIN time

twthorn:
We were using BEGIN timestamp for events. This does not match the [docs](DBZ-7628) which state that ts_ms is the time that the change is _applied_ to the database. This is also inconsistent with other connectors.

We add tests to reproduce the issue of using BEGIN timestamp. Then adjust the functionality to support using COMMIT timestamp instead.

One key part of the correctness of this code is that a the response.getEventsList() will only ever contain 1 transaction or a partial transaction. It will never contain two transactions (this is why we do not need to buffer/store multiple commit timestamps). We know this is the case by the [commit validation in the codebase](https://github.com/debezium/debezium-connector-vitess/blob/main/src/main/java/io/debezium/connector/vitess/connection/VitessReplicationConnection.java#L165-L167), ie that if another commit is ever received in a response.getEventsList() then we throw an exception.

### DIFF
--- a/src/main/java/io/debezium/connector/vitess/connection/MessageDecoder.java
+++ b/src/main/java/io/debezium/connector/vitess/connection/MessageDecoder.java
@@ -9,9 +9,13 @@ import io.debezium.connector.vitess.Vgtid;
 
 import binlogdata.Binlogdata;
 
+import java.time.Instant;
+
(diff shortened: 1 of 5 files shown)


jpechane:
@twthorn Applied, thanks!

PR #187, opened 2024-03-21
DBZ-7698 Add transaction order metadata

twthorn:
### Motivation
Provide transaction order metadata to establish ordering between any two operations for the same primary key, regardless of order consumed by the downstream application.

See [Zulip](https://debezium.zulipchat.com/#narrow/stream/348255-community-vitess/topic/Increasing.20Kafka.20Partitions) & [Jira](https://issues.redhat.com/browse/DBZ-7698) for more info on motivation.

### Implementation
The key new class introduced is OrderedTransactionContext. It is similar to TransactionContext but also adds two new fields, epoch & rank. Rank is a monotonically increasing numerical value that is used for comparing which transaction is more recent. Each database can have its own function for this, we given an example for Vitess. Epoch is a value that is incremented whenever this monotonically increasing property of rank would be violated (e.g., in Vitess, this would be shrinking the host set of the GTID, or a MySQL version upgrade). In order to prevent wrongfully incrementing the epoch, it must be stored in offsets state. We store the transaction rank alongside since they are semantically related (although not strictly necessary for transaction rank to be in offsets, similar to transaction order key).

### Additional Details
This demonstrates how this transaction order metadata is possible. Note: since this is to be a self-contained PR in the vitess connector repo, there are some workarounds that we do, that would be different if we want to more completely adopt this in the main Debezium repo. These are as follows:
1. OrderedTransactionContext subclasses TransactionContext - we do this just to plug in nicely with Debezium Core. What we should really do is either change TransactionContext to handle ordered metadata, or define an interface and have two classes one basic and one with ordered metadata.
2. EpochProvider/RankProvider interfaces - these should be in debezium-core's transaction package, and could be implemented by all connectors
3. The struct of a data change event is missing the epoch & rank information. This is needed in order for downstream applications to automatically handle repartition events. In order to add this, we must modify this [method](https://github.com/debezium/debezium/blob/main/debezium-core/src/main/java/io/debezium/pipeline/txmetadata/TransactionMonitor.java#L176-L182) (or create a way to modify the behavior when an ordered transaction context is presented) to include the new fields (we omit this for now since we are only making changes in the Vitess Connector repo).
4. Vgtid to per-shard gtid - Previously our transaction ID for metadata and other purposes was the full Vgtid (all shards & gtid sets). Now, with this change, in order to plug into the TransactionMonitor class & TransactionContext, we need to be able to deduce the rank & epoch solely from the transaction ID value. This means we cannot pass in Vgtid, because we would not know which shard to look at for the gtid set. Therefore we must pass in just the GTID for the desired shard. Note: if we change debezium-core we could avoid this (add some more functionalities for receiving shard info), but as of right now it works, and may be the preferred method (since we want debezium-core to remain generic (rather than customize to handle this gtid filtering for shard case) and the offsets we store still have vgtid which is needed to resuming a VStream).
5. More tests - for now there are some tests/assertions we can't do yet. Specifically for (3) once that is fixed we'd want to add assertions on the source info. Additionally, we should also have an integration test on the epoch increase case (we can manipulate the stored offsets to trigger epoch increase event like shrinking host set).

### DIFF
--- a/src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
+++ b/src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
@@ -389,6 +389,15 @@ public class VitessConnectorConfig extends RelationalDatabaseConnectorConfig {
                     + "'precise' represents values as precise (Java's 'BigDecimal') values;"
                     + "'long' represents values using Java's 'long', which may not offer the precision but will be far easier to use in consumers.");
 
+    public static final Field PROVIDE_ORDERED_TRANSACTION_METADATA = Field.create("provide.ordered.transaction.metadata")
+            .withDisplayName("Provide ordered transaction meatadata")
+            .withType(Type.BOOLEAN)
+            .withDefault(false)
+            .withWidth(Width.SHORT)
+            .withImportance(ConfigDef.Importance.HIGH)
+            .withDescription(
+                    "Whether to provided ordered metadata on transactions");
+
     public static final Field SOURCE_INFO_STRUCT_MAKER = CommonConnectorConfig.SOURCE_INFO_STRUCT_MAKER
             .withDefault(VitessSourceInfoStructMaker.class.getName());
 
@@ -417,7 +426,7 @@ public class VitessConnectorConfig extends RelationalDatabaseConnectorConfig {
             .events(
                     INCLUDE_UNKNOWN_DATATYPES,
                     SOURCE_INFO_STRUCT_MAKER)
-            .connector(SNAPSHOT_MODE, BIGINT_UNSIGNED_HANDLING_MODE)
+            .connector(SNAPSHOT_MODE, BIGINT_UNSIGNED_HANDLING_MODE, PROVIDE_ORDERED_TRANSACTION_METADATA)
             .excluding(SCHEMA_EXCLUDE_LIST, SCHEMA_INCLUDE_LIST)
             .create();
 
@@ -616,5 +625,9 @@ public class VitessConnectorConfig extends RelationalDatabaseConnectorConfig {
     public BigIntUnsignedHandlingMode getBigIntUnsgnedHandlingMode() {
         return BigIntUnsignedHandlingMode.parse(getConfig().getString(BIGINT_UNSIGNED_HANDLING_MODE),
                 BIGINT_UNSIGNED_HANDLING_MODE.defaultValueAsString());
+    }
+
+    public boolean shouldProvideOrderedTransactionMetadata() {
+        return Boolean.parseBoolean(getConfig().getString(PROVIDE_ORDERED_TRANSACTION_METADATA));
     }
 }
--- a/src/main/java/io/debezium/connector/vitess/transaction/Gtid.java
+++ b/src/main/java/io/debezium/connector/vitess/transaction/Gtid.java
@@ -1,0 +1,82 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.connector.vitess.transaction;
+
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+class Gtid {
+
+    public String getVersion() {
+        return version;
+    }
+
+    private String version = "";
+
+    public Set<String> getHosts() {
+        return hosts;
+    }
+
+    private Set<String> hosts = new HashSet();
+
+    public List<String> getSequenceValues() {
+        return sequenceValues;
+    }
+
+    private List<String> sequenceValues = new ArrayList();
+
+    private static final String PREFIX_LAST_CHAR = "/";
+
+    private static int getVersionEndIndex(String transactionId) {
+        return transactionId.indexOf(PREFIX_LAST_CHAR);
+    }
+
+    private static String trimVersion(String transactionId) {
+        int index = getVersionEndIndex(transactionId);
+        if (index != -1) {
+            return transactionId.substring(index + 1);
+        }
+        return transactionId;
+    }
+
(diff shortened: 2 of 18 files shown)


src/main/java/io/debezium/connector/vitess/transaction/RankProvider.java
```diff
@@ -0,0 +1,12 @@
...
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.connector.vitess.transaction;
+
+import java.math.BigInteger;
+
+interface RankProvider {
+    BigInteger getRank(String transactionId);
+}
```
Naros, line 12: Just thinking aloud; what about expanding this to allow arbitrary attributes somehow?  I'm just not sure that only a transaction Id would be sufficient.  For example, in Informix, if you don't have concurrent transactions, the underlying LSN doesn't change.  Just curious if passing a parameter object here would be more beneficial for more corner cases where a single `String` may not be sufficient.

twthorn: Responded in [Zulip](https://debezium.zulipchat.com/#narrow/stream/348255-community-vitess/topic/Increasing.20Kafka.20Partitions/near/428410735) overall we could use the source info struct to allow for all needed params to be passed to each connector.

src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
```diff
@@ -389,6 +389,15 @@ public static BigIntUnsignedHandlingMode parse(String value, String defaultValue
...
                     + "'long' represents values using Java's 'long', which may not offer the precision but will be far easier to use in consumers.");
 
+    public static final Field PROVIDE_ORDERED_TRANSACTION_METADATA = Field.create("provide.ordered.transaction.metadata")
+            .withDisplayName("Provide ordered transaction meatadata")
+            .withType(Type.BOOLEAN)
+            .withDefault(false)
+            .withWidth(Width.SHORT)
+            .withImportance(ConfigDef.Importance.HIGH)
+            .withDescription(
+                    "Whether to provided ordered metadata on transactions");
```
HenryCaiHaiying, line 399: `provided` --> `provide`

src/main/java/io/debezium/connector/vitess/transaction/Gtid.java
```diff
@@ -0,0 +1,82 @@
...
+
+    private String version = "";
+
+    public Set<String> getHosts() {
+        return hosts;
+    }
+
+    private Set<String> hosts = new HashSet();
+
+    public List<String> getSequenceValues() {
```
HenryCaiHaiying, line 27: It would be good to give an example of `transactionId` and the expected value of `version`, `sequenceValues`

HenryCaiHaiying (commented):
What is currently stored in the Kafka offset topic?  The VGTID or GTID?, where do you store the epoch information in Kafka offset topic?

src/main/java/io/debezium/connector/vitess/transaction/OrderedTransactionContext.java
```diff
@@ -0,0 +1,164 @@
...
+            Map.Entry<String, Long> e = (Map.Entry) var3.next();
+            offset.put(OFFSET_TABLE_COUNT_PREFIX + e.getKey(), e.getValue());
+        }
+        return offset;
+    }
+
+    public static OrderedTransactionContext load(Map<String, ?> offsets, EpochProvider epochProvider, RankProvider rankProvider) {
+        OrderedTransactionContext context = new OrderedTransactionContext(epochProvider, rankProvider);
+        context.transactionId = (String) offsets.get(OFFSET_TRANSACTION_ID);
+        context.previousTransactionId = (String) offsets.get(OFFSET_TRANSACTION_ID);
```
HenryCaiHaiying, line 77: `transactionId` and `previousTransactionId` is using the same key in the `offsets` map?

HenryCaiHaiying: The offset_transaction_id stored in offset map is for a specific shard.  A VGTID in the same offset map represents multiple shards.  If the offset_transaction_id in the offset map represents shard1, but the new events coming from VSTREAM is for shard2, how do you do the comparison?

twthorn: Ah yes forgot to add this part, we do need mapping of shard to epoch.

jpechane:
Two initial comments

* Should not `io.debezium.connector.vitess.SourceInfo` implement `io.debezium.connector.AbstractSourceInfo.sequence()` in this PR?
* I wonder if we need rank and epoch provider. If we are oing to have this really generic then maybe just a tuple of values should be used. And it is up to each connector to implement 1, 2, 3 or arbitrary values needed. WDYT?

twthorn:
> Should not io.debezium.connector.vitess.SourceInfo implement io.debezium.connector.AbstractSourceInfo.sequence() in this PR?

Transaction rank could be added here. Epoch couldn't because it needs state. However, I don't think we want to provide transaction rank for all events going forward (regardless of ordered transaction metadata setting). It isn't useful without epoch data.

For connectors that do have this, e.g., [postgres](https://github.com/debezium/debezium/blob/main/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/SourceInfo.java#L176-L193), it is metdata that is tied to each event, and not calcualted/derived, e.g., lsn.

> I wonder if we need rank and epoch provider. If we are oing to have this really generic then maybe just a tuple of values should be used. And it is up to each connector to implement 1, 2, 3 or arbitrary values needed. WDYT?

I agree with this, we want as general a solution as possible. I am working on implementation in debezium core that will allow for any values to be augmented to the transaction context. It could be tuple/array/etc., entirely up to the connector that implements it. The core code won't assume anything about rank/epoch.

PR #190, opened 2024-04-12
DBZ-7773 Implement time.precision.mode for adaptive_time_microseconds and connect

twthorn:
### Changes

Match what the [MySQL](https://debezium.io/documentation/reference/stable/connectors/mysql.html#mysql-temporal-types) connector does for handling temporal types with the time.precision.mode setting. Previously was converting everything to strings.

### Verificaiton
Add unit tests for the custom logic for conversion we added. Also add integration test for a table with varying precision levels to test how the behavior changes (e.g., micros or millis based on precision).


jpechane:
@twthorn Applied, thanks!

PR #194, opened 2024-05-28
DBZ-6721 & DBZ-6722 Add transforms: local vgtid, remove field, and filter transaction topic

twthorn:
[DBZ-6721](https://issues.redhat.com/browse/DBZ-6721)
[DBZ-6722](https://issues.redhat.com/browse/DBZ-6722)

Add three new transforms

### Filter Transaction Topic
For connectors where the volume is high (transaction topic is massive), we can drop all messages matching the schema

### Remove Field
Generic config for specifying field names (separated by dot '.' for nesting) to remove fields. Can be used to drop any field, in this case we use for transaction.id which can be very large in vitess (contains all shards gtid).

### Local VGTID
Update VGTID field to only contain vgtid for the shard that the change happened in.

We add this local vgtid transformation (add a hashmap so we don't have to iterate over all shards to do this).

We need to keep the vgtid global for proper offset/state storage so therefore it must be done only at the end as SMT before publishing the message (and does not affect source offset of the message).

### Itests
Also add itests for all these functionalities together for how they can be used to receive transaction order metadata without massively increasing data volume produced.

### DIFF
--- a/src/main/java/io/debezium/connector/vitess/Vgtid.java
+++ b/src/main/java/io/debezium/connector/vitess/Vgtid.java
@@ -6,6 +6,7 @@
 package io.debezium.connector.vitess;
 
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Objects;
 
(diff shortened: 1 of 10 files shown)


twthorn:
@jpechane all updated with refactor to move from core to here, tests are all passing, ready for review

jpechane:
@twthorn LGTM, thanks. Could you please rebase the PR on the latest main and remove merge commit(s)? When cleaned up the PR is ready to be merged.

twthorn:
@jpechane updated!

jpechane:
@twthorn Applied, thanks

jpechane:
Could you please provide a short update of vitess connector docs describing the new capability? Thanks

PR #196, opened 2024-05-29
DBZ-7905 Handle single position with host in gtid

twthorn:
We can have an edge case where a mysql/vitess host is only used for a single binlog position so it's not a range format host1:1-5 instead just host1:5 so we adapt parsing logic to handle this and reproduce with a test.


twthorn:
@jpechane this one is also ready for review, thanks!

jpechane:
@twthorn Applied, thanks

PR #197, opened 2024-05-30
DBZ-7910 Fix column exclude filter type, add itest

twthorn:
We were previously wrongly using catalog (which stores the shard as done in https://github.com/debezium/debezium-connector-vitess/pull/149) for doing column exclusion. Change to schema (keyspace) and add itest to reproduce issue.


jpechane:
@twthorn Applied, thanks

PR #199, opened 2024-06-18
DBZ-7962 Add heartbeats

twthorn:
### Summary

In the case of low volume write tables, we need to be able to emit heartbeat events.

### Design

Other connectors do this with a heartbeat query that is run regularly in the upstream database. We don't need to do this since Vitess/VStream provides a helpful flag called [HeartbeatInterval](https://vitess.io/docs/17.0/reference/vreplication/vstream/#heartbeatinterval). We can use the HeartbeatInterval config already present in Debezium core to get this value, parse as seconds, and pass to the VStream API. With this set up, the vstream channel will receive regular heartbeat events from the vtgate indicating liveness. We can dispatch downstream heartbeat events for each we receive.

In order for the heartbeat to be useful, we must know which vitess shards it pertains to (e.g., 8 shards, 2 tasks, each streaming data for 4 shards, if one task is down, how do we detect it from the heartbeats?). So we include the vgtid in each heartbeat's payload, this vgtid forms the collections of shards for this task and will allow us to determine which shards this heartbeat is indicating liveness for.

### Implementation
Various changes to support another data type Heartbeat. Handling it correctly in the replication connection (ie send the event even if we didn't receive a complete transaction, ie between transactions). Decode and read the timestamp properly. Call dispatch heartbeat event for each heartbeat event we receive. We call it roughly as frequently as its own heartbeat expiration logic (assuming no events to send, if there are events it will be longer, which is fine since heartbeats are needed when there are no events, and we want to keep up with realtime rather than be processing heartbeat events if there are data events to process).

Note: there is an edge case where there can be many events sent (commit/begin) but no data changes for the filtered tables. In this case, no data change records are published to the data change topic & Vitess VStream will not send a heartbeat event because there is plenty of events being sent over the vstream, so no records are sent to the heartbeat topic. This would wrongfully mean no heartbeat events are emitted or a live/healthy connector. To counteract this, also dispatch heartbeat events on COMMIT events. If there is a surplus of data not for this table, it means that there are other tables receiving many transactions so these events will be received frequently enough. HeartbeatImpl superclass uses its own rate limiting so its fine if we call more often than necessary.

### Testing 
Set up some itests to verify we get heartbeat events when no other writes are happening. Other tests to verify handling of heartbeat events.

### DIFF
--- a/src/test/java/io/debezium/connector/vitess/VitessConnectorIT.java
+++ b/src/test/java/io/debezium/connector/vitess/VitessConnectorIT.java
@@ -151,12 +151,12 @@ public class VitessConnectorIT extends AbstractVitessConnectorTest {
     }
 
     @Test
-    @FixFor("DBZ-2776")
+    @FixFor("DBZ-7962")
     public void shouldReceiveHeartbeatEvents() throws Exception {
         TestHelper.executeDDL("vitess_create_tables.ddl");
         startConnector(config -> config.with(
                 Heartbeat.HEARTBEAT_INTERVAL.name(), 1000),
-                true);
+                false);
         assertConnectorIsRunning();
 
         String topic = Heartbeat.HEARTBEAT_TOPICS_PREFIX.defaultValueAsString() + "." + TEST_SERVER;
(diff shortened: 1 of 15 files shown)


twthorn:
@jpechane can you review when you get a chance? thanks!

src/test/java/io/debezium/connector/vitess/VitessConnectorIT.java
```diff
@@ -168,6 +168,31 @@ public void shouldReceiveHeartbeatEvents() throws Exception {
...
+        startConnector(config -> config.with(
+                Heartbeat.HEARTBEAT_INTERVAL.name(), 1000).with(
+                        VitessConnectorConfig.SHARD, expectedShard),
+                true);
+        assertConnectorIsRunning();
+
+        String topic = Heartbeat.HEARTBEAT_TOPICS_PREFIX.defaultValueAsString() + "." + TEST_SERVER;
+        int expectedHeartbeatRecords = 1;
+        // Sleep for 3 seconds, heartbeat sent every 1 second
+        Thread.sleep(3 * 1000);
```
jpechane, line 185: The tests based on sleep tends to be brittle. Could you please modify it so use `Awaitility` waiting for a message arrival.

jpechane (commented):
I left a comment about testing. Otherwise LGTM!

jpechane:
@twthorn Applied, thanks!

PR #202, opened 2024-07-11
DBZ-8053 Handle empty shards

twthorn:
[DBZ-8053](https://issues.redhat.com/browse/DBZ-8053)

In some cases, a Vitess keyspace can have empty shards (shards without tablets). Do the following to handle this:
1. Provide a config exclude.empty.shards (boolean) to enable this behavior
2. If enabled, and no shard list is specified, then auto detect the non-empty shards (shards with tablets)
3. If enabled, and a shard list is specified, then route metadata queries (e.g., get tables) only to those listed shards. Although we could filter this to find the non-empty shards, we want to stay true to what the user configured (and fail loudly if a user is expecting to have a vstream for a certain set of shards, and some of those are empty). However, we can still route the queries such that they use the configured shards and do not risk failing because there are empty shards not in the list.
4. For the detected non-empty shard list, use it for query routing and for setting up the vstreams.
5. Refactor our metadata queries (eg getting shards/tables) into one class and do this empty shard handling there.

### DIFF
--- a/src/main/java/io/debezium/connector/vitess/VitessMetadata.java
+++ b/src/main/java/io/debezium/connector/vitess/VitessMetadata.java
@@ -79,7 +79,7 @@ public class VitessMetadata {
     }
 
     private static void logResponse(Vtgate.ExecuteResponse response, String query) {
-        LOGGER.info("Got response: {} for query: {}", response, query);
+        LOGGER.debug("Got response: {} for query: {}", response, query);
     }
 
     private List<String> getVitessShards() {
--- a/src/main/java/io/debezium/connector/vitess/VitessMetadata.java
+++ b/src/main/java/io/debezium/connector/vitess/VitessMetadata.java
@@ -31,21 +31,26 @@ import io.vitess.proto.Vtgate;
 public class VitessMetadata {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(VitessMetadata.class);
+    private VitessConnectorConfig config;
 
-    public static List<String> getShards(VitessConnectorConfig config) {
+    public VitessMetadata(VitessConnectorConfig config) {
+        this.config = config;
+    }
+
+    public List<String> getShards() {
         List<String> shards;
         if (config.excludeEmptyShards()) {
             LOGGER.info("Excluding empty shards");
-            shards = getVitessShardsFromTablets(config);
+            shards = getVitessShardsFromTablets();
         }
         else {
-            shards = getVitessShards(config);
+            shards = getVitessShards();
         }
         LOGGER.info("Shards: {}", shards);
         return shards;
     }
 
-    public static List<String> getTables(VitessConnectorConfig config) {
+    public List<String> getTables() {
         Vtgate.ExecuteResponse response;
         String query;
         if (config.excludeEmptyShards()) {
@@ -57,15 +62,15 @@ public class VitessMetadata {
             }
             else {
                 LOGGER.info("Getting tables from a non-empty shard");
-                shardsToQuery = getVitessShardsFromTablets(config);
+                shardsToQuery = getVitessShardsFromTablets();
             }
             String randomShard = shardsToQuery.get(new Random().nextInt(shardsToQuery.size()));
             LOGGER.info("Get tables from shard: {}", randomShard);
-            response = executeQuery(config, query, randomShard);
+            response = executeQuery(query, randomShard);
         }
         else {
             query = String.format("SHOW TABLES FROM %s", config.getKeyspace());
-            response = executeQuery(config, query);
+            response = executeQuery(query);
         }
         logResponse(response, query);
         List<String> tables = getFlattenedRowsFromResponse(response);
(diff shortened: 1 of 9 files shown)


twthorn:
@jpechane can you take a look when you get the chance? thanks!

src/main/java/io/debezium/connector/vitess/VitessMetadata.java
```diff
@@ -0,0 +1,196 @@
...
+
+import io.debezium.annotation.VisibleForTesting;
+import io.debezium.connector.vitess.connection.VitessReplicationConnection;
+import io.vitess.proto.Query;
+import io.vitess.proto.Vtgate;
+
+/**
+ * Class for getting metadata on Vitess, e.g., tables, shards. Supports shard-specific queries.
+ */
+public class VitessMetadata {
```
jpechane, line 31: Would it be possible to convert this class to be no-static? A lot of methods uses `VitessConnectorConfig` so maybe an instance can be created that will receive the config in constructor.

src/main/java/io/debezium/connector/vitess/connection/VitessReplicationConnection.java
```diff
@@ -75,6 +76,21 @@ public Vtgate.ExecuteResponse execute(String sqlStatement) {
         return newBlockingStub(channel).execute(request);
     }
 
+    public Vtgate.ExecuteResponse execute(String sqlStatement, String shard) {
+        LOGGER.info("Executing sqlStament {}", sqlStatement);
```
jpechane, line 80:
```suggestion
        LOGGER.debug("Executing sqlStament {}", sqlStatement);
```

src/main/java/io/debezium/connector/vitess/connection/VitessReplicationConnection.java
```diff
@@ -75,6 +76,21 @@ public Vtgate.ExecuteResponse execute(String sqlStatement) {
...
+        ManagedChannel channel = newChannel(config.getVtgateHost(), config.getVtgatePort(), config.getGrpcMaxInboundMessageSize());
+        managedChannel.compareAndSet(null, channel);
+
+        String target = String.format("%s:%s@%s", config.getKeyspace(), shard, config.getTabletType());
+        Vtgate.Session session = Vtgate.Session.newBuilder().setTargetString(target).setAutocommit(true).build();
+        LOGGER.info("Autocommit {}", session.getAutocommit());
```
jpechane, line 86:
```suggestion
        LOGGER.debug("Autocommit {}", session.getAutocommit());
```

src/test/java/io/debezium/connector/vitess/VitessConnectorIT.java
```diff
@@ -1025,6 +1028,152 @@ public void shouldUseSameTransactionIdWhenMultiGrpcResponses() throws Exception
         Testing.Print.enable();
     }
 
+    @Test
```
jpechane, line 1031: Would it be possible to add a test to verify the behaviour when `EXCLUDE_EMPTY_SHARDS` is not set?

src/main/java/io/debezium/connector/vitess/VitessMetadata.java
```diff
@@ -0,0 +1,201 @@
...
+            response = executeQuery(query);
+        }
+        logResponse(response, query);
+        List<String> tables = getFlattenedRowsFromResponse(response);
+        LOGGER.info("All tables from keyspace {} are: {}", config.getKeyspace(), tables);
+        return tables;
+    }
+
+    private static void logResponse(Vtgate.ExecuteResponse response, String query) {
+        LOGGER.info("Got response: {} for query: {}", response, query);
```
jpechane, line 82:
```suggestion
        LOGGER.debug("Got response: {} for query: {}", response, query);
```

src/main/java/io/debezium/connector/vitess/VitessMetadata.java
```diff
@@ -0,0 +1,201 @@
...
+                response = connection.execute(query);
+            }
+            return response;
+        }
+        catch (Exception e) {
+            throw new RuntimeException(String.format("Unexpected error while running query: %s", query), e);
+        }
+    }
+
+    private static List<String> getFlattenedRowsFromResponse(Vtgate.ExecuteResponse response) {
```
jpechane, line 129: Is it possible to make the rest of the methods non-static too?

jpechane (commented):
@twthorn LGTM, thanks! I left just a couple of minor comments.

src/main/java/io/debezium/connector/vitess/VitessMetadata.java
twthorn: It is possible, but I don't think we want to, since they don't rely on any instance variables. Static clearly conveys these methods do not rely on object state.

src/test/java/io/debezium/connector/vitess/VitessConnectorIT.java
twthorn: If `EXCLUDE_EMPTY_SHARDS` is not set, and we try to stream from the keyspace with an empty shard, it fails and brings down the test vitess instance, so subsequent unrelated tests would also fail. All other tests in the integration tests do not set this variable and it defaults to false so they are effectively testing it without EXCLUDE_EMPTY_SHARDS being set.

jpechane:
@twthorn Applied, thanks!

PR #204, opened 2024-07-22
DBZ-8087 Get previous vgtid from vgtid key in offsets

twthorn:
It is non-deterministic/race condition whether or not the `OFFSET_TRANSACTION_ID` is present in the offsets (if we happen to store offsets during a transaction, [it is added since it's not null, otherwise it is not added](https://github.com/debezium/debezium/blob/main/debezium-core/src/main/java/io/debezium/pipeline/txmetadata/TransactionContext.java#L51)). For example if heartbeats are enabled, there is not transaction but we still commit offsets.

Epoch needs to be based on the vgtid we store as for the VGTID_KEY since that is always present in offsets regardless of whether we are in a transaction or not. This fixes an edge case where we saw epoch being set back down to zero because we were getting null for our previous VGTID string. Also fail loudly if we ever see this case again (null previous VGTID but there is state for the epoch, this is invalid state and should throw exception).

### DIFF
--- a/src/main/java/io/debezium/connector/vitess/pipeline/txmetadata/VitessEpochProvider.java
+++ b/src/main/java/io/debezium/connector/vitess/pipeline/txmetadata/VitessEpochProvider.java
@@ -70,16 +70,22 @@ public class VitessEpochProvider {
             }
         }
         catch (JsonProcessingException e) {
-            throw new RuntimeException("Cannot read epoch: " + shardToEpoch.toString());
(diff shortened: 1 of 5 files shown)


HenryCaiHaiying (approved):
lgtm

jpechane (changes requested):
@twthorn LGTM, thanks! I left just a minor comment around code structure.

twthorn:
@jpechane this is ready for re-review, thanks!

jpechane:
@twthorn Applied, thanks!

PR #207, opened 2024-08-15
DBZ-8154 Make order metadata epoch handle changes to task parallelism & shard set

twthorn:
Right now each task simply reads the epoch from its last stored offsets. As seen by the logic for vgtid, we need to be able to handle changes in generation if task offset storage mode is enabled (e.g., task parallelism change or shard subset change). Add this feature by refactoring the logic for vgtid to be more generic so we can do equivalent operations on the shard epochs (when enabled). Add `ShardEpochMap` for centralizing the ser/deser operations of reading the shard epoch map from state & updating it. Operate on VGTID and ShardEpochMap in similar manners to accomplish the same goal of redistributing correctly on task parallelism or shard change.

Note: there is one "TODO" intentionally left in the code: we will soon put out another PR to establish shard lineage and allow for inheriting the epoch of a parent shard when a split occurs. This PR is already substantial so we will do that in a separate one.

### DIFF
--- a/src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
+++ b/src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
@@ -31,6 +31,7 @@ import io.debezium.config.Field;
 import io.debezium.config.Field.ValidationOutput;
 import io.debezium.connector.SourceInfoStructMaker;
 import io.debezium.connector.vitess.connection.VitessTabletType;
+import io.debezium.connector.vitess.pipeline.txmetadata.ShardEpochMap;
 import io.debezium.heartbeat.Heartbeat;
 import io.debezium.heartbeat.HeartbeatConnectionProvider;
 import io.debezium.heartbeat.HeartbeatErrorHandler;
@@ -247,6 +248,17 @@ public class VitessConnectorConfig extends RelationalDatabaseConnectorConfig {
                             + " It has to be set together with vitess.shard."
                             + " If not configured, the connector streams changes from the latest position for the given shard(s)."
                             + " If snapshot.mode is INITIAL (default), the connector starts copying the tables for the given shard(s) first regardless of gtid value.");
+
+    public static final Field SHARD_EPOCH_MAP = Field.create(VITESS_CONFIG_GROUP_PREFIX + "shard.epoch.map")
+            .withDisplayName("shard.epoch.map")
+            .withType(Type.STRING)
+            .withWidth(Width.LONG)
+            .withDefault("")
+            .withImportance(ConfigDef.Importance.LOW)
+            .withValidation(VitessConnectorConfig::validateShardEpochMap)
+            .withDescription(
+                    "ShardEpochMap to use for the initial epoch values for the given shards. If not configured the connector streams changes" +
+                            "from a default value of 0.");
 
     public static final Field GTID = Field.create(VITESS_CONFIG_GROUP_PREFIX + "gtid")
             .withDisplayName("gtid")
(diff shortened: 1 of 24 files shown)


twthorn:
@jpechane Can you review this when you get the chance? Thanks!

src/main/java/io/debezium/connector/vitess/VitessConnectorTask.java
```diff
@@ -148,52 +150,104 @@ protected ChangeEventSourceCoordinator<VitessPartition, VitessOffsetContext> sta
...
-                    gtidStr = prevGtidsPerShard.get(shard);
-                    LOGGER.warn("Using offsets from previous gen: shard {}, gen {}, gtid {}", shard, gen, gtidStr);
-                }
-                else {
-                    gtidStr = configGtidsPerShard.getOrDefault(shard, null);
-                    LOGGER.warn("Using offsets from config: shard {}, gtid {}", shard, gtidStr);
-                }
-                shardGtids.add(new Vgtid.ShardGtid(keyspace, shard, gtidStr));
+            config = getVitessTaskValuePerShard(config, connectorConfig, VitessOffsetRetriever.ValueType.GTID);
+            if (VitessOffsetRetriever.isShardEpochMapEnabled(connectorConfig)) {
```
HenryCaiHaiying, line 157: This looks like should be if ... else structure otherwise line 156 was executed in vain.

src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
```diff
@@ -248,6 +249,17 @@ public static BigIntUnsignedHandlingMode parse(String value, String defaultValue
                             + " If not configured, the connector streams changes from the latest position for the given shard(s)."
                             + " If snapshot.mode is INITIAL (default), the connector starts copying the tables for the given shard(s) first regardless of gtid value.");
 
+    public static final Field SHARD_EPOCH_MAP = Field.create(VITESS_CONFIG_GROUP_PREFIX + "shard.epoch.map")
```
HenryCaiHaiying, line 252: This is going to be a big map to pass in and potentially error-prone, when will people pass in this epoch on command line?   It this config mainly used for carry the current epoch values from a running system?

HenryCaiHaiying (commented):
It seems to me that the change is mainly about introducing the concept of ShardEpoch map and refactoring code to share the implementation of shardEpoch and shardVGtid, the title of the PR might need to reflect that.

src/main/java/io/debezium/connector/vitess/VitessConnectorTask.java
twthorn: The `getVitessTaskValuePerShard` actually modifies the map (since it passes it in as an arg). However, that is not clear (not well named) and better to not have hidden side effects on the config. I updated so it simply returns the value and then the config editing is done in this function.

twthorn (commented):
> It seems to me that the change is mainly about introducing the concept of ShardEpoch map and refactoring code to share the implementation of shardEpoch and shardVGtid, the title of the PR might need to reflect that.

That is all true. The result is that now, similar to VGTID, it will be redistributed correctly whenever the shards assigned to tasks changes (eg task parallelism change or shard set change). So the title highlights the effect, not so much the implementation. I added more details on what you mentioned to the description to help make it clearer.

HenryCaiHaiying (approved):
lgtm, thanks for the refactoring and support for shardEpoch

jpechane:
@twthorn Looks nice, thanks!

jpechane:
Please just send a docs PR with the new config option.

PR #208, opened 2024-08-21
DBZ-8163 Add inherit epoch feature

twthorn:
Add a config feature for inherit epoch. When enabled, the parent shards will be deduced from the hex ranges and we will take the max of the parent shards + 1 to get a newer epoch value. Refactor some things to allow us to read in this config. Also refactor the epoch logic so that our shard epoch map does not drift from the vgtid we receive.

I am hoping to expand our integration test suite in a separate PR to test reshard operations. Since this PR is already substantial putting this out now and will have a separate one for adding reshards to itests.

### DIFF
--- a/src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
+++ b/src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
@@ -32,6 +32,7 @@ import io.debezium.config.Field.ValidationOutput;
 import io.debezium.connector.SourceInfoStructMaker;
 import io.debezium.connector.vitess.connection.VitessTabletType;
 import io.debezium.connector.vitess.pipeline.txmetadata.ShardEpochMap;
+import io.debezium.connector.vitess.pipeline.txmetadata.VitessOrderedTransactionMetadataFactory;
 import io.debezium.heartbeat.Heartbeat;
 import io.debezium.heartbeat.HeartbeatConnectionProvider;
 import io.debezium.heartbeat.HeartbeatErrorHandler;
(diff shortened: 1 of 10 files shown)


src/main/java/io/debezium/connector/vitess/pipeline/txmetadata/ShardLineage.java
```diff
@@ -0,0 +1,95 @@
...
+        }
+
+        private void validateBounds() {
+            if (this.lowerBound.compareTo(this.upperBound) >= 0) {
+                throw new IllegalArgumentException("Invalid shard range " + this);
+            }
+        }
+
+        public boolean overlaps(Shard shard) {
+            return this.lowerBound.compareTo(shard.upperBound) < 0 && this.upperBound.compareTo(shard.lowerBound) > 0;
```
HenryCaiHaiying, line 66: How about the situation when the two ranges meet on the boundary ( == 0 case)?

src/main/java/io/debezium/connector/vitess/pipeline/txmetadata/ShardLineage.java
```diff
@@ -0,0 +1,95 @@
...
+
+        return maxParentEpoch + 1;
+    }
+
+    private static class Shard {
+
+        // A string lexicographically less than all other strings
+        public static final String NEGATIVE_INFINITY = "";
+        // A string lexicographically greater than all other strings
+        public static final String POSITIVE_INFINITY = "\uFFFF";
```
HenryCaiHaiying, line 47: Is 4F long/big enough, do we need 8F?

HenryCaiHaiying (commented):
Looks good in general, left a few small comments.

src/main/java/io/debezium/connector/vitess/pipeline/txmetadata/ShardLineage.java
twthorn: Vitess ranges are have inclusive lower bounds and exclusive upper bounds so this should be correct ([docs](https://vitess.io/docs/19.0/reference/features/sharding/#shard-names)). We can think about each part:
1. `this.lowerBound.compareTo(shard.upperBound) == 0` the lower bound is the same as the upper bound we are comparing to. Since the upper bound is exclusive, this interval does not overlap.
2. `this.upperBound.compareTo(shard.lowerBound) == 0` the upper bound is the same as the lower bound we are comparing to. Since the upper bound is exclusive, these two intervals do not overlap.

I added a test to demonstrate this case.

src/main/java/io/debezium/connector/vitess/pipeline/txmetadata/ShardLineage.java
twthorn: Ah this `\uFFFF` is actually a single special char that is greater than all other chars. I used the Character package to make this clear (it is the MAX_VALUE). And it turns out there is MIN_VALUE which is another special char (more clear than empty string). I updated the comments to describe this too.

HenryCaiHaiying (approved):
Thanks for the explanation, the change looks good to me.

src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
```diff
@@ -301,6 +302,15 @@ public static BigIntUnsignedHandlingMode parse(String value, String defaultValue
             .withDescription("Control StopOnReshard VStream flag."
                     + " If set true, the old VStream will be stopped after a reshard operation.");
 
+    public static final Field INHERIT_EPOCH = Field.create(VITESS_CONFIG_GROUP_PREFIX + "inherit_epoch")
```
jpechane, line 305:
```suggestion
    public static final Field INHERIT_EPOCH = Field.create(VITESS_CONFIG_GROUP_PREFIX + "inherit.epoch")
```

jpechane:
@twthorn Applied, thanks!

PR #209, opened 2024-09-12
DBZ-8235 Add invalid value logger to VitessValueConverter

twthorn:
Similar to https://github.com/debezium/debezium/pull/5507 this can be logged for every record so users may want to suppress it without losing log values for the other log messages so add a separate logger.


jpechane:
@twthorn Applied, thanks

PR #213, opened 2024-10-31
DBZ-8354 Fix table filter logic for vstream

twthorn:
The previous code had custom regex filtering which meant it would match all tables with suffixes for the table filter (eg filter is keyspace.table1 and would also match keyspace.table1_suffix1, keyspace.table1_suffix2, keyspace.table1_suffix3). These events would later be filtered out with the call to emit data change event which uses the Debezium Filters construct correctly (it only matches the full table name). This means there could be lots of unnecessary events being requested & deserialized from the vstream. This PR removes this custom regex logic and instead use the Debezium `Filters` to determine which tables to filter fixing this edge case.

This should significantly help with CPU & performance (and cost savings) for keyspaces that have many tables that have the same prefix and different suffixes.

### DIFF
--- a/src/main/java/io/debezium/connector/vitess/VitessConnector.java
+++ b/src/main/java/io/debezium/connector/vitess/VitessConnector.java
@@ -14,7 +14,6 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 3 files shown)


jpechane:
@twthorn Applied, thanks!

PR #215, opened 2024-11-07
DBZ-8325 Emit DDL events

twthorn:
Based on feedback on https://github.com/debezium/debezium-connector-vitess/pull/210 creating this PR. Depends on https://github.com/debezium/debezium/pull/5979 (so tests won't pass until that is merged) which allows for publishing schema change events without using the Historized configs/schemas. These are not necessary to use for Vitess VStream so it simplifies the code greatly.

Note: although Vitess DDLs use MySQL syntax, there is a requirement to use the MySQL parser that the [databaseTables](https://github.com/debezium/debezium/blob/44251bc442a8481c8ae3ee9bf70f06b405f328d2/debezium-ddl-parser/src/main/java/io/debezium/antlr/AntlrDdlParser.java#L62) be up to date prior to receiving a DDL event otherwise it is trivially ignored (eg [here](https://github.com/debezium/debezium/blob/44251bc442a8481c8ae3ee9bf70f06b405f328d2/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/antlr/listener/AlterTableParserListener.java#L50)). Since Vitess VStream may send a DDL as the first event (with no prior FIELD event to establish the table schema), then this is not feasible to know (without doing a full historized snapshot with CREATE table statements). To work around this we use a simple regex and don't bother parsing the full SQL text. This still allows us to set fields correctly (eg `ddl` and `table` field are correct, and we simply leave `tableChanges` field empty). In a separate ticket we can change the parser to possibly support this, but that is a larger change/refactor that I believe would best be in a separate PR.

Example DDL event:
```
SourceRecord{sourcePartition={server=test_server}, sourceOffset={vgtid=[{"keyspace":"test_unsharded_keyspace","shard":"0","gtid":"MySQL56/7359ecd4-9d4c-11ef-971d-0242ac110002:1-1783","table_p_ks":[]}]}} ConnectRecord{topic='test_server', kafkaPartition=0, key=Struct{databaseName=test_unsharded_keyspace}, keySchema=Schema{io.debezium.connector.vitess.SchemaChangeKey:STRUCT}, value=Struct{source=Struct{version=3.0.0-SNAPSHOT,connector=vitess,name=test_server,ts_ms=1731014782000,db=,ts_us=1731014782000000,ts_ns=1731014782000000000,keyspace=test_unsharded_keyspace,table=ddl_table,shard=0,vgtid=[{"keyspace":"test_unsharded_keyspace","shard":"0","gtid":"MySQL56/7359ecd4-9d4c-11ef-971d-0242ac110002:1-1784","table_p_ks":[]}]},ts_ms=1731014782527,databaseName=test_unsharded_keyspace,ddl=alter table ddl_table add column new_column_name INT,tableChanges=[]}, valueSchema=Schema{io.debezium.connector.vitess.SchemaChangeValue:STRUCT}, timestamp=null, headers=ConnectHeaders(headers=)}
```

### DIFF
--- a/src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
+++ b/src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
@@ -348,6 +348,20 @@ public class VitessConnectorConfig extends RelationalDatabaseConnectorConfig {
                             + "'false' (the default) omits the fields; "
                             + "'true' converts the field into an implementation dependent binary representation.");
 
+    // Needed for backward compatibility, otherwise all upgraded connectors will start publishing schema change events
+    // by default.
+    public static final Field INCLUDE_SCHEMA_CHANGES = Field.create("include.schema.changes")
+            .withDisplayName("Include database schema changes")
+            .withType(Type.BOOLEAN)
+            .withGroup(Field.createGroupEntry(Field.Group.CONNECTOR, 0))
+            .withWidth(Width.SHORT)
+            .withImportance(ConfigDef.Importance.MEDIUM)
+            .withDescription("Whether the connector should publish changes in the database schema to a Kafka topic with "
+                    + "the same name as the database server ID. Each schema change will be recorded using a key that "
+                    + "contains the database name and whose value include logical description of the new schema and optionally the DDL statement(s). "
+                    + "The default is 'true'. This is independent of how the connector internally records database schema history.")
+            .withDefault(false);
+
     public static final Field OFFSET_STORAGE_PER_TASK = Field.create(VITESS_CONFIG_GROUP_PREFIX + "offset.storage.per.task")
             .withDisplayName("Store offsets per task")
             .withType(Type.BOOLEAN)
@@ -367,6 +381,14 @@ public class VitessConnectorConfig extends RelationalDatabaseConnectorConfig {
             .withImportance(ConfigDef.Importance.LOW)
             .withValidation(CommonConnectorConfig::validateTopicName)
             .withDescription("Overrides the topic.prefix used for the data change topic.");
+
+    public static final Field OVERRIDE_SCHEMA_CHANGE_TOPIC = Field.create("override.schema.change.topic")
+            .withDisplayName("Override schema change topic name")
+            .withType(Type.STRING)
+            .withWidth(Width.MEDIUM)
+            .withImportance(ConfigDef.Importance.LOW)
+            .withValidation(CommonConnectorConfig::validateTopicName)
+            .withDescription("Overrides the name of the schema change topic (if not set uses topic.prefx).");
 
     public static final Field OFFSET_STORAGE_TASK_KEY_GEN = Field.create(VITESS_CONFIG_GROUP_PREFIX + "offset.storage.task.key.gen")
             .withDisplayName("Offset storage task key generation number")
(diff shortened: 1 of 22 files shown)


twthorn:
@jpechane new PR for debezium-vitess-connector is ready for review

src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
```diff
@@ -368,6 +382,14 @@ public static BigIntUnsignedHandlingMode parse(String value, String defaultValue
             .withValidation(CommonConnectorConfig::validateTopicName)
             .withDescription("Overrides the topic.prefix used for the data change topic.");
 
+    public static final Field OVERRIDE_SCHEMA_CHANGE_TOPIC = Field.create("override.schema.change.topic")
```
twthorn, line 385: @jpechane Following up on discussion [here](https://github.com/debezium/debezium-connector-vitess/pull/210/files#r1825801913), we do still need this config. The [SchemaTopicNamingStrategy](https://github.com/debezium/debezium/blob/main/debezium-core/src/main/java/io/debezium/schema/SchemaTopicNamingStrategy.java#L40-L42) does not override `schemaChangeTopic` so it defaults to the implementation in [AbstractTopicNamingStrategy](https://github.com/debezium/debezium/blob/main/debezium-core/src/main/java/io/debezium/schema/AbstractTopicNamingStrategy.java#L124) ie just using topic.prefix. We do need a custom one. I updated the docstring in the `vitess.TableTopicNamingStrategy` to provide more context on this (ie with vitess-connectors we can have mulitple connectors for the same database, so need to be able to override this).

jpechane: You provide `io.debezium.connector.vitess.TableTopicNamingStrategy`. So why can't you provide two strategies that would differ in shcema handling? Or if you want to make it configurable, there is a `configure` method at teh strategy class level so you can move the config field into the strategy itself in the same way as is done at `io.debezium.schema.AbstractTopicNamingStrategy`. That way it will be configurable yet the field will not be defined at connector level.

jpechane (commented):
@twthorn When we solve the issue of the topic strategy configuration the PR is good to go. It seems its quite isolated now and feature is disabled by default so I think it can go even into 3.0.x. I plan to do 3.0.2 on Thu/Fri so if we can sort it out it an be merged and released this week.

src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
twthorn: If we separate the strategies (one does the override data change topic, one does the schema change topic override) then afaik we can't combine two topic naming strategies in the config so we cannot use both of them. If vitess-connectors want to enable this feature and they have multiple connectors per database they would need to use both (see this [test](https://github.com/debezium/debezium-connector-vitess/pull/215/files#diff-1b9011e9a90b9250c8885c36b838ac867f597b3726cf037ec1fa0aefb47d8221R264-R265) for demonstration).

I will go with that second option of moving the configs into the vitess.TableTopicNamingStrategy class.

jpechane:
@twthorn And it is applied! Thanks for the effort and patience, it is very nice result.

PR #216, opened 2024-11-20
DBZ-8438 Pin netty versions to prevent data corruption

twthorn:
4.1.111 has data corruption see [here](https://debezium.zulipchat.com/#narrow/channel/348255-community-vitess/topic/Data.20Corruption.20with.20Kafka.203.2E9.2E0/near/483564599)

I could pin this in debezium-core, but I figure Debezium may want to keep the versions aligned with kafka for those that do not use grpc-java package. Any other connectors that use grpc-java should have this fix as well.


jpechane:
@twthorn Applied, thanks!

PR #217, opened 2024-12-02
DBZ-8479 Improve DDL parsing logic

twthorn:
A few changes to make this DDL logic more fault tolerant
1. Only parse DDLs if that schema history is enabled (otherwise this is unnecessary work)
2. Strip comments to ensure we parse correctly
3. Make table & type fault tolerant so we gracefully handle cases where we didn't parse either

Also fix [DBZ-8432](https://issues.redhat.com/browse/DBZ-8432)

Also dockerfile failing with the apt-key statement so remove that as it's unnecessary


twthorn:
@jpechane Can you take a look when you get the chance? Thank you

jpechane:
@twthorn Applied, thanks!

PR #220, opened 2024-12-19
DBZ-8541 Handle partially overridden transactions

twthorn:
See ticket
https://issues.redhat.com/browse/DBZ-8541

If we have the task with shards
s1: current
s2: current

Then the first vgtid we will receive will be
s1: gtid1
s2: current

Then at some point we will get a transaction for s2 as well and will be
s1: gtid1
s2: gtid2

Right now we fail on this case even though it's valid.

Also do some miscellaneous improvements
- Simply vitess epoch provider logic
- Add tests for gtid, remove unnecessary methods
- Make validation & other checks more thorough
- Make handling more general so we fail less often
- Add more tests for above case and others


jpechane:
@twthorn Applied, thanks!

PR #222, opened 2025-01-07
DBZ-8561 Correctly handle enum/set types in Vitess v20+

twthorn:
See this [message](https://debezium.zulipchat.com/#narrow/channel/348255-community-vitess/topic/VStream.20ENUM.20.2F.20SET.20Changes/near/435005995) for core logic used. We read this new flag that is set by newer vitess versions to indicate if the enum/set values are strings (previously only in snapshot, going forward is in snapshot & regular phase).

This makes Debezium forward compatible with Vitess upgrades to v20+ so we handle enum/set types correctly.

Without these changes, if https://github.com/debezium/debezium/pull/6079 is merged then tests trivially fail. With these changes and that other PR then tests pass.

Right now, since the main upgrade isn't merged, tests trivially pass (does not use any of the new code)

### DIFF
--- a/src/main/java/io/debezium/connector/vitess/VitessType.java
+++ b/src/main/java/io/debezium/connector/vitess/VitessType.java
@@ -91,7 +91,7 @@ public class VitessType {
     }
 
     // Resolve JDBC type from vstream FIELD event
-    public static VitessType resolve(Query.Field field, boolean isInVStreamCopy) {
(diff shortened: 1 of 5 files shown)


twthorn:
@jpechane this is ready for review when you get the chance, thanks!

jpechane:
@twthorn Applied, thanks!

PR #224, opened 2025-01-21
DBZ-8594 Prevent data loss when primary key update is last operation in transaction

twthorn:
We found an edge case described in [DBZ-8594](https://issues.redhat.com/browse/DBZ-8594)

There's another possible way to implement this: Modify VitessChangeRecordEmitter - we could modify our implementation to set the offset before emitting the final delete record [here](https://github.com/debezium/debezium/blob/main/debezium-core/src/main/java/io/debezium/relational/RelationalChangeRecordEmitter.java#L184). This would be possible but it seemed like a more custom solution, we'd prefer to be as standard as possible. Let me know if you think this is preferred. I think VitessOffsetContext would be called there to resetVgtid (and maybe in create/delete/update functions as well). The thing I don't like is we don't get the bug fixes or changes from the class we inherit from so we have to maintain more custom/overridden logic.

The main downside of this approach is seen in `testCopyTableAndRestart` ie if we do a special gtid (for snapshot or to speed up to current) then after that no other operations happen (no new vgtid's) then the restart vgtid is still the special one (snapshot or fastforward to current) so that operation would get repeated. Perhaps the gudiance we can give users is to enable transaction metadata (so begin/commit events are always sent even if tables are unaffected), but even that is not a guarantee.

Couple examples of other connectors I think may be at risk
[Postgres](https://github.com/debezium/debezium/blob/main/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresStreamingChangeEventSource.java#L328-L331)
[SQL Server](https://github.com/debezium/debezium/blob/main/debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerStreamingChangeEventSource.java#L329)

### DIFF
--- a/src/main/java/io/debezium/connector/vitess/VitessStreamingChangeEventSource.java
+++ b/src/main/java/io/debezium/connector/vitess/VitessStreamingChangeEventSource.java
@@ -94,16 +94,25 @@ public class VitessStreamingChangeEventSource implements StreamingChangeEventSou
 
     private ReplicationMessageProcessor newReplicationMessageProcessor(VitessPartition partition,
                                                                        VitessOffsetContext offsetContext) {
-        return (message, newVgtid, isLastRowOfTransaction) -> {
+        return (message, newVgtid) -> {
             if (message.isTransactionalMessage()) {
-                // Tx BEGIN/END event
-                offsetContext.rotateVgtid(newVgtid, message.getCommitTime());
+                // Tx BEGIN/COMMIT event
                 if (message.getOperation() == ReplicationMessage.Operation.BEGIN) {
+                    // When BEGIN event is received, newVgtid will have been populated with the transaction's vgtid, rotate
+                    // to set currentVgtid to newVgtid and restartVgtid to the previous transaction VGTID
+                    offsetContext.rotateVgtid(newVgtid, message.getCommitTime());
                     // send to transaction topic
                     VitessTransactionInfo transactionInfo = new VitessTransactionInfo(message.getTransactionId(), message.getShard());
                     dispatcher.dispatchTransactionStartedEvent(partition, transactionInfo, offsetContext, message.getCommitTime());
                 }
                 else if (message.getOperation() == ReplicationMessage.Operation.COMMIT) {
+                    // When COMMIT event is received, all events have been processed except for this COMMIT event
+                    // We reset the VGTID such that current & restart VGTIDs are equal to this transaction's VGTID
+                    // We send one final event (transaction committed), the offset will only be committed if that event
+                    // is sent successfully.
+                    // If transaction metadata is disabled, then the offset will not be updated until a message of the next
+                    // transaction is sent (next transaction's restartVgtid = this transaction's currentVgtid)
+                    offsetContext.resetVgtid(newVgtid, message.getCommitTime());
                     // send to transaction topic
                     dispatcher.dispatchTransactionCommittedEvent(partition, offsetContext, message.getCommitTime());
                     // Send a heartbeat event if time has elapsed
@@ -154,10 +163,6 @@ public class VitessStreamingChangeEventSource implements StreamingChangeEventSou
 
                 offsetContext.event(tableId, message.getCommitTime());
                 offsetContext.setShard(message.getShard());
-                if (isLastRowOfTransaction) {
-                    // Right before processing the last row, reset the previous offset to the new vgtid so the last row has the new vgtid as offset.
-                    offsetContext.resetVgtid(newVgtid, message.getCommitTime());
-                }
                 dispatcher.dispatchDataChangeEvent(
                         partition,
                         tableId,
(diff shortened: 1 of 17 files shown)


twthorn:
@jpechane Can you give this is a look when you get the chance? Thanks!

HenryCaiHaiying (commented):
Is that possible to use Kafka transaction to solve the problem?

If DELETE, Tombstone, CREATE are 3 events belong to the same Vitess Transaction, should we put the three message publishing into Kafka in the same Kafka Transaction?  The downstream consumer won't see the messages until the whole transaction commits?

twthorn:
@HenryCaiHaiying There are a few reasons to avoid transactions
- Tight coupling - Debezium embedded engine mode allows Debezium to be run in a standard java application without kafka/kafka connect. We do not want to make any assumptions about what other external systems will be associated with Debezium. We simply want each record to contain correct content (ie the offset refers to the last fully processed transaction)
- Performance - enabling transactions has some throughput hit
- Rigidity - all deploys must have transactions enabled to operate correctly

src/main/java/io/debezium/connector/vitess/VitessStreamingChangeEventSource.java
```diff
@@ -94,16 +94,25 @@ public void execute(ChangeEventSourceContext context, VitessPartition partition,
 
     private ReplicationMessageProcessor newReplicationMessageProcessor(VitessPartition partition,
                                                                        VitessOffsetContext offsetContext) {
-        return (message, newVgtid, isLastRowOfTransaction) -> {
+        return (message, newVgtid) -> {
             if (message.isTransactionalMessage()) {
-                // Tx BEGIN/END event
-                offsetContext.rotateVgtid(newVgtid, message.getCommitTime());
```
HenryCaiHaiying, line 100: In the old code, rotateVgtid() is called on all cases of isTransactionMessage() (including BEGIN and COMMIT and other unknown types), do we need to cover that unknown types in the new code?

src/main/java/io/debezium/connector/vitess/VitessStreamingChangeEventSource.java
```diff
@@ -154,10 +163,6 @@ else if (message.getOperation().equals(ReplicationMessage.Operation.HEARTBEAT))
 
                 offsetContext.event(tableId, message.getCommitTime());
                 offsetContext.setShard(message.getShard());
-                if (isLastRowOfTransaction) {
-                    // Right before processing the last row, reset the previous offset to the new vgtid so the last row has the new vgtid as offset.
-                    offsetContext.resetVgtid(newVgtid, message.getCommitTime());
```
HenryCaiHaiying, line 159: So we lost this resetVgid() in the new code, I guess you are assuming the COMMIT message will follow later, does COMMIT message always happens?   Also from the old commit, it seems the last row has the new vgtid, looks like this won't be the same in the new code, any implications on this behavior change?

twthorn: > does COMMIT message always happens?

VitessReplicationConnection ensures that we always receive a COMMIT after a BEGIN (otherwise throws an error). There is one vstream copy [edge case](https://github.com/debezium/debezium-connector-vitess/blob/main/src/main/java/io/debezium/connector/vitess/connection/VitessReplicationConnection.java#L162) where duplicate BEGINs can be received, but it discards the events so this code doesn't need to handle those.

> Also from the old commit, it seems the last row has the new vgtid, looks like this won't be the same in the new code, any implications on this behavior change?

Yes, this is the key part that leads to the data loss bug.

Let the current transaction VGTID be `n`, previous is `n-1`, next transaction is `n+1`

Previous behavior:
1. BEGIN - rotateVgtid - set currentVgtid to `n`, restartVgtid is `n-1`
2. INSERT/UPDATE/DELETE not last operation - no-op on offset context
3. INSERT/UPDATE/DELETE that is the last operation - reset vgtid - sets the currentVgtid to be equal to the `n`. Set restartVgtid equal to `n` (wrongfully, since it may still have multiple messages to send in case of primary key update (delete, tombstone, create), and can fail part way between them).

New behavior:
1. BEGIN - rotateVgtid - set currentVgtid to `n`, restartVgtid is the `n-1`
2. INSERT/UPDATE/DELETE any order, last or not - no op on offset context
3. COMMIT - resetVgtid - set current & restart VGTIDs to `n` - the offset will only be committed if the commit event is successfully produced (if tx metadata is enabled) or when it receives a write for the `n+1` (since the offset vgtid ie the restartVgtid will point to `n`) or if it sends a heartbeat event (in the case no tx metadata, and no subsequent writes, that heartbeat will have offset / restartVgtid set to `n`)

src/main/java/io/debezium/connector/vitess/VitessStreamingChangeEventSource.java
HenryCaiHaiying: It might be safer to add an else {assert here} in case our assumptions are not correct.

src/main/java/io/debezium/connector/vitess/VitessStreamingChangeEventSource.java
HenryCaiHaiying: Sounds good

src/main/java/io/debezium/connector/vitess/VitessStreamingChangeEventSource.java
twthorn: I added validation when transactional messages are created that they are only commit or begin. Changed to else since these are the only two cases. And add unit tests for verifying this begin/commit validation

HenryCaiHaiying (approved):
lgtm

twthorn:
@jpechane all feedback addressed, if you can take a look when you get the chance. Thanks!

jpechane:
@twthorn Applied, thanks!

PR #226, opened 2025-02-04
DBZ-8663 Modify itest to test truncation for all string data types

twthorn:
We didn't have full test coverage for all string data types, in particular one missed was json. Verify truncation for all string data types


jpechane:
@twthorn Applied, thanks

PR #230, opened 2025-02-20
DBZ-8719 Test scaling down reads last offsets

twthorn:
https://issues.redhat.com/browse/DBZ-8719

Add tests to verify that when we scale down tasks that we read the last stored offsets (we already have tests that verify this for the scale out case).


twthorn:
@jpechane this is ready for review when you get the chance, thank you!

jpechane:
@twthorn Applied, thanks

PR #231, opened 2025-03-05
Improve performance by avoiding repeated operations

twthorn:
[DBZ-8757](https://issues.redhat.com/browse/DBZ-8757)

After some [tuning](https://debezium.zulipchat.com/#narrow/channel/302529-community-general/topic/Performance.20Tuning.20with.20Full.20Queue.20but.20Underfilled.20Batches) we found that the queue was consistently empty. After profiling there were wasted CPU cycles repeating these operations.
1. VitessChangeRecordEmitter - Getting config/table schema can be done once rather than in each before/after call.
2. Vgtid - there is no way to modify the Vgtid so we can instead cache the string representation rather than deserializing to json string each time.

These changes ~2x'd performance and now the queue is full, so we are now enqueueing data fast enough to exceed dequeue speed.

### DIFF
--- a/src/main/java/io/debezium/connector/vitess/Vgtid.java
+++ b/src/main/java/io/debezium/connector/vitess/Vgtid.java
@@ -6,10 +6,11 @@
 package io.debezium.connector.vitess;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.Objects;
-import java.util.Optional;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonIgnore;
@@ -35,13 +36,15 @@ public class Vgtid {
 
     @JsonIgnore
     private final Binlogdata.VGtid rawVgtid;
-    private final List<ShardGtid> shardGtids = new ArrayList<>();
-    private final HashMap<String, ShardGtid> shardNameToShardGtid = new HashMap<>();
-    // Vgtid cannot be modified, so we can cache the string representation to improve performance
-    private Optional<String> cachedVgtidString = Optional.empty();
+    private final List<ShardGtid> shardGtids;
+    private final Map<String, ShardGtid> shardNameToShardGtid;
+    // Vgtid cannot be modified, so we store the string representation to improve performance
+    private final String vgtidString;
 
     private Vgtid(Binlogdata.VGtid rawVgtid) {
         this.rawVgtid = rawVgtid;
+        List<Vgtid.ShardGtid> shardGtids = new ArrayList<>();
+        HashMap<String, ShardGtid> shardNameToShardGtid = new HashMap<>();
         for (Binlogdata.ShardGtid shardGtid : rawVgtid.getShardGtidsList()) {
             TablePrimaryKeys tabletablePrimaryKeys = new TablePrimaryKeys(shardGtid.getTablePKsList());
             ShardGtid currentShardGtid = new ShardGtid(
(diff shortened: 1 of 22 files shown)


jpechane (commented):
@twthorn Thanks for the PR. Please create a Jira for that. The changes definitely makes sense. I just proposed an alternative for `Vgtid`.

jpechane:
@twthorn Any comment the the `Vgtid`? Also I see the PR comment refers to one Jira and commit comment to another.

twthorn:
Yes I'll update this PR soon. I agree with the feedback, we can set it eagerly.

twthorn:
Vgtids are very large so it is often beneficial to remove them to improve performance. However, the previous provided transform recreates the entire schema/struct. We change this to just set to empty string which leads to another 1.5x improvement. Also do some other misc improvements to reduce redundant operations/calls.

src/main/java/io/debezium/connector/vitess/Vgtid.java
```diff
@@ -34,25 +36,37 @@ public class Vgtid {
...
-    private final HashMap<String, ShardGtid> shardNameToShardGtid = new HashMap<>();
+    private final List<ShardGtid> shardGtids;
+    private final Map<String, ShardGtid> shardNameToShardGtid;
+    // Vgtid cannot be modified, so we store the string representation to improve performance
+    private final String vgtidString;
 
     private Vgtid(Binlogdata.VGtid rawVgtid) {
         this.rawVgtid = rawVgtid;
+        List<Vgtid.ShardGtid> shardGtids = new ArrayList<>();
+        HashMap<String, ShardGtid> shardNameToShardGtid = new HashMap<>();
```
jpechane, line 47:
```suggestion
        Map<String, ShardGtid> shardNameToShardGtid = new HashMap<>();
```

src/main/java/io/debezium/connector/vitess/Vgtid.java
```diff
@@ -34,25 +36,37 @@ public class Vgtid {
...
+        this.vgtidString = shardGtidsToString(shardGtids);
     }
 
-    private Vgtid(List<ShardGtid> shardGtids) {
-        this.shardGtids.addAll(shardGtids);
+    private Vgtid(List<ShardGtid> shardGtids, String vgtidString) {
+        this.vgtidString = vgtidString;
+        this.shardGtids = Collections.unmodifiableList(shardGtids);
+
+        HashMap<String, ShardGtid> shardNameToShardGtid = new HashMap<>();
```
jpechane, line 65:
```suggestion
        Map<String, ShardGtid> shardNameToShardGtid = new HashMap<>();
```

src/main/java/io/debezium/connector/vitess/transforms/ReplaceFieldValue.java
```diff
@@ -0,0 +1,129 @@
...
+import org.apache.kafka.connect.data.Schema;
+import org.apache.kafka.connect.data.Struct;
+import org.apache.kafka.connect.transforms.Transformation;
+
+import io.debezium.Module;
+import io.debezium.config.Configuration;
+import io.debezium.config.Field;
+import io.debezium.transforms.SmtManager;
+
+/**
```
jpechane, line 25: Please add a JavaDoc with comprehensive dscription.

src/main/java/io/debezium/connector/vitess/transforms/ReplaceFieldValue.java
```diff
@@ -0,0 +1,129 @@
...
+ * @author Thomas Thornton
+ */
+public class ReplaceFieldValue<R extends ConnectRecord<R>> implements Transformation<R>, Versioned {
+    private static final String FIELD_DELIMITER = ",";
+
+    public static final String FIELD_NAMES_CONF = "field_names";
+
+    public static final Field FIELD_NAMES_FIELD = Field.create(FIELD_NAMES_CONF)
+            .withDisplayName("List of field names to remove, full path eg source.database or transaction.id")
```
jpechane, line 34: remove or replace?

src/test/java/io/debezium/connector/vitess/transforms/ReplaceFieldValueTest.java
```diff
@@ -0,0 +1,90 @@
...
+ * @author Thomas Thornton
+ */
+public class ReplaceFieldValueTest {
+
+    private static final String TRANSACTION_BLOCK_SCHEMA_NAME = "event.block";
+    public static final int TRANSACTION_BLOCK_SCHEMA_VERSION = 1;
+
+    @Test
+    public void shouldReplaceTxIdWhenEnabled() {
+        ReplaceFieldValue<SourceRecord> removeFields = new ReplaceFieldValue<SourceRecord>();
```
jpechane, line 31: `removeFields` or `replaceFields`? The same issue is present in other methods too.

jpechane (changes requested):
@twthorn Thanks. I pushed an import fix to get tests running.
Still I had to leave few comments, mostly about documentation and naming - might related to copy and paste.

twthorn:
Several of the changes are not needed to achieve the new max throughput of 60k msg/sec. So I undid those

jpechane:
@twthorn Applied, thanks! Please just create PR with docs for the new SMT.

twthorn:
@jpechane Will do! Thank you for fixing the description

twthorn:
https://github.com/debezium/debezium/pull/6250

PR #232, opened 2025-03-12
DBZ-8775 Add support for new VStream feature: streamKeyspaceHeartbeats

twthorn:
[DBZ-8775](https://issues.redhat.com/browse/DBZ-8775)

https://github.com/debezium/debezium/pull/6224 must first be merged (tests will fail trivially until then, once it's merged all tests will pass)

This was added in [Vitess v21](https://github.com/vitessio/vitess/issues/16477).

Add support to pass in this VStream API flag.

Since these heartbeat tables are regular data change events, add a config that allows user to pass in a regex exclude list for the override data change topic config (ie multiple connectors in the same DB streaming the same heartbeat table should output to separate topics).


twthorn:
@jpechane can you review this and also https://github.com/debezium/debezium/pull/6224 when you get the chance? Thanks!

jpechane (commented):
@twthorn LGTM, thanks! This can be merged when CI passes.

jpechane (approved):

PR #233, opened 2025-03-18
DBZ-8824 Add integration test for keyspace heartbeats during snapshot

twthorn:
[DBZ-8824](https://issues.redhat.com/browse/DBZ-8824)

Adds an integration test for keyspace heartbeats during a copy. In particular we need to verify that heartbeat events are not sent during the [catch up phase](https://github.com/vitessio/vitess/issues/6277) of a vstream copy. This is so that downstream users do not wrongfully trigger jobs during a snapshot (thinking data is all caught up).

A few other changes
1. Remove unused variables
2. DRY up some code
3. Extend some test utilities
4. Isolate heartbeat keyspace


twthorn:
@jpechane When you get the chance, this is ready for review. Thanks!

jpechane:
@twthorn Applied, thanks!

PR #234, opened 2025-03-19
DBZ-8826 Support temporal precision mode for strings

twthorn:
[DBZ-8826](https://issues.redhat.com/browse/DBZ-8826)

We should also be able to output as strings for temporal types (time, date, datetime). Accomplish this by using the TemporalPrecisionMode.ISO_STRING mode. Note: In order to prevent loss of information, we customize the behavior compared to other connectors. Specifically, we want to ensure that impossible values like 0000-00-00 are still sent identical to their upstream values. Other connectors convert to a timestamp then back to a string (so converting 0000-00-00 fails). Instead, we do not do this conversion, and simply send the string value sent by MySQL.

Note: later we could modify debezium core to have a mode called RAW_STRING that does this. I went with this approach to keep things contained with vitess-connector and to not block on that. Later, If it's something we think other connectors want & we add it to core, then we can refactor the vitess-connector to use it.

Other changes
- Remove some code no longer needed
- Refactor some tests

### DIFF
--- a/src/test/java/io/debezium/connector/vitess/VitessValueConverterTest.java
+++ b/src/test/java/io/debezium/connector/vitess/VitessValueConverterTest.java
@@ -19,10 +19,11 @@ import java.util.List;
 import org.apache.kafka.connect.data.Field;
 import org.apache.kafka.connect.data.Schema;
 import org.apache.kafka.connect.data.SchemaBuilder;
-import org.junit.Before;
+import org.junit.After;
 import org.junit.Test;
 
 import io.debezium.connector.vitess.connection.VStreamOutputMessageDecoder;
+import io.debezium.jdbc.TemporalPrecisionMode;
 import io.debezium.junit.logging.LogInterceptor;
 import io.debezium.relational.Column;
 import io.debezium.relational.Table;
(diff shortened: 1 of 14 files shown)


twthorn:
@jpechane This is also ready for review. Thank you for the help on these!

src/test/java/io/debezium/connector/vitess/VitessValueConverterTest.java
```diff
@@ -61,11 +59,19 @@ public void before() {
         decoder = new VStreamOutputMessageDecoder(schema);
     }
 
+    @After
+    public void before() {
```
jpechane, line 63: `before` ro `@After`? I'd say the naming is a bit confusing :-)

src/test/java/io/debezium/connector/vitess/VitessValueConverterTest.java
```diff
@@ -117,8 +124,24 @@ public void shouldGetInt64SchemaBuilderForTime() throws InterruptedException {
...
+        for (String columnName : List.of("date_col", "time_col", "datetime_col")) {
+            Column column = table.columnWithName(columnName);
+            SchemaBuilder schemaBuilder = converter.schemaBuilder(column);
+            assertThat(schemaBuilder.schema()).isEqualTo(Schema.STRING_SCHEMA);
+        }
+    }
+
     @Test
     public void shouldGetYearSchemaBuilderForYear() throws InterruptedException {
+        setUpResources(TestHelper.defaultConnectorConfig());
```
jpechane, line 144: Maybe instead of adding the explicit calls you could provide a dedicated test class that would test temporal string conversion only?

jpechane (changes requested):
@twthorn LGTM, thanks! I left two comments related to test structure. Other than that we are good to go.

src/test/java/io/debezium/connector/vitess/VitessValueConverterTest.java
twthorn: Nice catch, updated!

src/test/java/io/debezium/connector/vitess/VitessValueConverterTest.java
twthorn: Great idea, updated!

jpechane:
@twthorn Applied, thanks!

PR #235, opened 2025-03-31
DBZ-8861 Add workload name to all queries issued by Debezium

twthorn:
This is supported by [Vitess](https://vitess.io/docs/21.0/user-guides/configuration-advanced/comment-directives/#workload-name-workload_name) for tracking which application is issuing the queries. Add this workload name info in every query we use (refactor things so all queries are issued from VitessMetadata).

Also clarify one test I noticed can be stronger assertion (assert epoch incremented for all shards).


twthorn:
@jpechane when you get the chance, this is ready for review, thanks!

jpechane:
@twthorn Applied, thanks! We have something similar for PostgreSQL and it can be handy.

PR #237, opened 2025-05-05
DBZ-9008 Log stack trace when connection fails

twthorn:
https://issues.redhat.com/browse/DBZ-9008

We were omitting the stack trace, use logger.error to include it as throwable arg.


Naros (approved):
Thanks @twthorn

PR #238, opened 2025-05-06
DBZ-9014 Make load balancing policy configurable

twthorn:
[DBZ-9014](https://issues.redhat.com/browse/DBZ-9014)

Make load balancing policy configurable.

The default is pick first (what it does currently when not specified). Round robin is also supported. Others may be added as well, so future proof it by keeping the config open as a string (with validation that the policy is supported for the current version).

### DIFF
--- a/src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
+++ b/src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
@@ -42,6 +42,9 @@ import io.debezium.relational.ColumnFilterMode;
 import io.debezium.relational.RelationalDatabaseConnectorConfig;
 import io.debezium.schema.SchemaNameAdjuster;
 import io.debezium.spi.topic.TopicNamingStrategy;
+import io.grpc.LoadBalancerProvider;
+import io.grpc.LoadBalancerRegistry;
+import io.grpc.internal.GrpcUtil;
 
 /**
  * Vitess connector configuration, including its specific configurations and the common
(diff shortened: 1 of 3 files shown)


src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
```diff
@@ -346,6 +349,15 @@ public static BigIntUnsignedHandlingMode parse(String value, String defaultValue
             .withValidation(Field::isInteger)
             .withDescription("Specify the maximum message size in bytes allowed to be received on the channel.");
 
+    public static final Field GRPC_DEFAULT_LOAD_BALANCING_POLICY = Field.create(VITESS_CONFIG_GROUP_PREFIX + "grpc.default_load_balancing_policy")
```
jpechane, line 352:
```suggestion
    public static final Field GRPC_DEFAULT_LOAD_BALANCING_POLICY = Field.create(VITESS_CONFIG_GROUP_PREFIX + "grpc.default.load.balancing.policy")
```

src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
```diff
@@ -346,6 +349,15 @@ public static BigIntUnsignedHandlingMode parse(String value, String defaultValue
...
+            .withDisplayName("VStream gRPC default load balancing policy")
+            .withType(Type.STRING)
+            .withWidth(Width.MEDIUM)
+            .withDefault(GrpcUtil.DEFAULT_LB_POLICY)
+            .withImportance(ConfigDef.Importance.MEDIUM)
+            .withValidation(VitessConnectorConfig::validateLoadBalancingPolicy)
+            .withDescription("Specify the maximum message size in bytes allowed to be received on the channel.");
```
jpechane, line 359: Description is invalid. Also it should probably contain at least two examples of the valid values

jpechane (commented):
@twthorn LGTM, thanks! I left request for two minor corrections, other than that it is good to go.

src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
twthorn: Nice catch, updated!

jpechane:
@twthorn Applied, thanks!

PR #242, opened 2025-06-11
DBZ-9110 Align to core changes

mfvitale:
requires: https://github.com/debezium/debezium/pull/6496


mfvitale:
Failure seems not related.

PR #246, opened 2025-07-08
DBZ 9176 Remove heartbeat creation from the configuration in favor of using `HeartbeatFactory`

kmos:
## Context

Actually in Debezium there are different way to instantiate an Heartbeat, which are not aligned.

## Changes
we propose to remove the usage in the `ConnectorConfiguration` in favor of `HeartBeatFactory`. Furthermore we deprecate the `constructors` in favor of a parameter based `factory`. 

## References
#depends https://github.com/debezium/debezium/pull/6530


mfvitale (approved):

Naros (approved):

Naros:
Applied, thanks @kmos !

PR #268, opened 2026-01-10
Support VStream transaction chunking (prevents out of memory errors)

### LINKED ISSUE #1519: Support for transaction chunking in vitess-connector
## Feature request or enhancement

Add support for transaction chunking flag to prevent VTGate OOMs

We found a critical vitess bug that can cause OOMs. The fix is opt in by vstream clients so debezium needs a config to allow users to set this
https://github.com/vitessio/vitess/pull/18849 

---

**Implementation ideas (optional)**

Will put out a PR

twthorn:
## Description

Fixes https://github.com/debezium/dbz/issues/1519

Note: This PR will fail trivially until the vitess-grpc-client version is released & upgraded.

Add support for transaction chunking

## Testing
Integration tests
1. Verify that Debezium receives a write operation distributed across shards with chunking enabled as two atomic chunked transactions on each shard (not interleaved). And verifies that total order is tracked correctly across chunked transactions
2. (Currently will fail) verify that enabling/migrating to chunking, ensures that the transaction ordering guarantees are preserved

###  TODO

https://github.com/debezium/dbz/issues/1530
The second itest is intentionally failing and is from TDD to demonstrate the edge case of switching from using the current VGTID to trailing/last VGTID (only feasible one to send when chunking is enabled). I will add another PR that adds a feature called connector generation, which can be used to trigger an epoch bump when switching on chunking. Since epoch is the first variable compared between events, it will allow all events with trailing/last VGTID after chunking is enabled to still be evaluated as newer for downstream consumers. See [here](https://debezium.io/documentation/reference/stable/connectors/vitess.html#vitess-ordered-transaction-metadata) for more details on comparison logic:
> 1. If the values for transaction_epoch are not equal, return the event with the higher transaction_epoch value. Otherwise, continue.
> 2. If the values for transaction_rank are not equal, return the event with the higher transaction_rank value. Otherwise, continue.
> 3. Return the event with a greater total_order value.

### DIFF
--- a/src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
+++ b/src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
@@ -337,6 +337,19 @@ public class VitessConnectorConfig extends RelationalDatabaseConnectorConfig {
             .withDescription("Control the interval between periodic gPRC keepalive pings for VStream." +
                     " Defaults to Long.MAX_VALUE (disabled).");
 
+    public static final Field TRANSACTION_CHUNK_SIZE_BYTES = Field.create(VITESS_CONFIG_GROUP_PREFIX + "transaction.chunk.size.bytes")
+            .withDisplayName("Transaction chunk size (bytes)")
+            .withType(Type.LONG)
+            .withDefault(0L)
+            .withWidth(Width.SHORT)
+            .withImportance(ConfigDef.Importance.HIGH)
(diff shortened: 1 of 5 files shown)


jpechane (commented):
@twthorn LGTM, thanks! Could you please fix the commit message and sign off the commit?

twthorn:
Blocked on this release https://github.com/vitessio/vitess-releaser/issues/140

PR #282, opened 2026-05-05
debezium/dbz#1887 Bug fix for copy completed sequence event & null last pk

twthorn:
We add unit test to repdouce that null past pk case.

For the event order it's non-deterministic due to vitess/grpc batching. We observed in our prod deployment and such a change resolved it there. Longer term we can refactor to make the logic unit-testable and we could add a test for that case I opened that for future here https://github.com/debezium/dbz/issues/1888


twthorn:
@jpechane when you get a free moment, this is ready for review

jpechane:
@twthorn Applied, thanks

PR #285, opened 2026-05-18
debezium/dbz#1938 Use new QueueProviderService API

Naros:
Fixes debezium/dbz#1938


Naros:
Test failures appear to be a small disconnect between the core repo `dbz#1938` and the cross-maven build with the new `EnforceRecordSize` behavior. As those are the only failures, I'm going to consider this green.

PR #286, opened 2026-06-10
debezium/dbz#2054 Backtick-quote keyspace and table identifiers in Vitess SQL

mcrauwel:
Fixes debezium/dbz#2054.

## Summary

Vitess keyspace and table names that aren't valid bare MySQL identifiers
(most commonly names containing a hyphen) caused metadata queries to fail
at vtgate with a syntax error, after which the connector silently returned
zero tables and produced no events.

Three sites interpolated names into SQL without quoting/escaping:
- `VitessMetadata#getTables` — `SHOW TABLES FROM <keyspace>`
- `VitessMetadata#getVitessShards` — `SHOW VITESS_SHARDS LIKE '<keyspace>/%'`
  (also did not escape LIKE wildcards, so an unescaped `_` in a keyspace
  name would over-match)
- `VitessReplicationConnection` VStream filter SQL — `select * from \`<table>\``
  (did not escape embedded backticks in a table name)

Adds `quoteIdentifier`, `escapeStringLiteral`, and `escapeLikePattern`
helpers on `VitessMetadata` and routes every interpolation site through
them.

## Test plan

- New unit tests in `VitessMetadataTest` cover plain identifiers,
  hyphenated names, embedded backticks, and LIKE wildcards including the
  ordering invariant (backslash escaped before `_`/`%`).
- DCO sign-off present.

### DIFF
--- a/src/main/java/io/debezium/connector/vitess/VitessMetadata.java
+++ b/src/main/java/io/debezium/connector/vitess/VitessMetadata.java
@@ -46,6 +46,28 @@ public class VitessMetadata {
         return formattedQuery;
     }
 
+    // MySQL identifier quoting: wrap in backticks and escape any embedded backtick by doubling.
+    // Required because keyspace/table names can legally contain characters that are not valid
+    // in bare MySQL identifiers (e.g. hyphens).
+    public static String quoteIdentifier(String identifier) {
+        return "`" + identifier.replace("`", "``") + "`";
+    }
+
+    // Escape a value to be placed inside a MySQL single-quoted string literal.
+    @VisibleForTesting
+    static String escapeStringLiteral(String value) {
+        return value.replace("\\", "\\\\").replace("'", "\\'");
+    }
+
+    // Escape LIKE wildcards so they match literally. Backslash is the LIKE escape character and
+    // must be escaped first; otherwise an unescaped `_` in a keyspace name would over-match
+    // (e.g. `foo_bar` would also match `foo-bar`). Apply before escapeStringLiteral so the added
+    // backslashes are themselves doubled when embedded in the literal.
+    @VisibleForTesting
+    static String escapeLikePattern(String value) {
+        return value.replace("\\", "\\\\").replace("_", "\\_").replace("%", "\\%");
+    }
+
     public List<String> getShards() {
         List<String> shards;
         if (config.excludeEmptyShards()) {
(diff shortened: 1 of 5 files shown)


src/main/java/io/debezium/connector/vitess/VitessMetadata.java
```diff
@@ -46,6 +46,28 @@ protected static String formatQuery(String query, String... args) {
         return formattedQuery;
     }
 
+    // MySQL identifier quoting: wrap in backticks and escape any embedded backtick by doubling.
+    // Required because keyspace/table names can legally contain characters that are not valid
+    // in bare MySQL identifiers (e.g. hyphens).
+    public static String quoteIdentifier(String identifier) {
```
jpechane: I don't think this should belong here. We have `io.debezium.jdbc.JdbcConnection.quoteIdentifier(String)` so IMHO the method should be a non-static one in `VitessReplicationConnection`.

jpechane (commented):
@mcrauwel Thanks for the PR. I proposed a small change in the implementation. Also it would be great to have an integration test that uses the table name with the problematic chars like dash and backtick.
One more question, is not there a similar issue for column names?

src/main/java/io/debezium/connector/vitess/VitessMetadata.java
mcrauwel: Good point, thanks — moved `quoteIdentifier` to a non-static method on `VitessReplicationConnection` in a2ad0e9, mirroring `JdbcConnection#quoteIdentifier`. `VitessMetadata#getTables` now obtains the quoted keyspace through the connection it opens for the query. The `escapeStringLiteral`/`escapeLikePattern` helpers stayed on `VitessMetadata` since they escape string literals/LIKE patterns for the `SHOW VITESS_SHARDS` query rather than identifiers — happy to move those too if you prefer.

mcrauwel:
Thanks for the review @jpechane!

**Integration test:** added `shouldStreamFromTableWithDashInName` in a2ad0e9, which creates a `dash-table` table, puts it in `table.include.list`, and verifies a change event streams through. For backticks it turns out Vitess itself cannot stream such tables at all: with a correctly quoted filter (`` select * from `backtick``table` ``) the vstreamer fails server-side with ``vttablet: ... unknown table backtick`table in schema`` — the tablet schema engine does not track backticked table names. So backtick escaping is covered by unit tests (`VitessReplicationConnectionTest`, `VitessMetadataTest`), but an end-to-end test would fail inside Vitess regardless of what the connector does.

**Column names:** I checked all SQL the connector builds — it is only the three sites touched by this PR (`SHOW TABLES FROM`, `SHOW VITESS_SHARDS LIKE`, and the VStream filter, which is always `select * from <table>` with no column list). Column names flow from VStream FIELD events and DDL messages into the connector but are never interpolated back into queries, so they are not affected.

jpechane:
@mcrauwel Applied, thanks!

PR #287, opened 2026-06-12
debezium/dbz#2060 Honor time.precision.mode=connect for TIMESTAMP columns

mcrauwel:
Fixes https://github.com/debezium/dbz/issues/2060

MySQL `TIMESTAMP` is mapped to `Types.TIMESTAMP_WITH_TIMEZONE`, for which `JdbcValueConverters` hardcodes `ZonedTimestamp` regardless of the configured `time.precision.mode`, so `connect` mode was silently ignored for `TIMESTAMP` columns while being honored for `DATETIME`, `DATE`, and `TIME`.

This change intercepts the `TIMESTAMP` path in `VitessValueConverter` when the mode is `connect` and emits Kafka Connect's `Timestamp` logical type (int64 epoch millis), parsing the UTC string emitted by VStream. The MySQL zero-date sentinel (`0000-00-00 00:00:00`) follows the same path as a `DATETIME` zero-date: null for optional columns, the schema default or epoch for non-optional columns.

**Scope / compatibility notes:**
* Only `connect` mode changes. All other precision modes (`adaptive_time_microseconds`, `isostring`) keep emitting `ZonedTimestamp` for `TIMESTAMP`, matching the documented exclusion. The unchanged expectations of the existing non-connect integration tests cover this.
* For users running `time.precision.mode=connect` with `TIMESTAMP` columns this is a behavior change (string `ZonedTimestamp` → int64 `Timestamp`) — i.e. the documented contract is now honored. May be worth a release-note mention.
* The existing connect-mode integration test expectations codified the old behavior and are updated accordingly; new unit and integration tests cover the conversion including zero-date values on both nullable and non-nullable columns.

A matching documentation update for `vitess.adoc` (adding the `TIMESTAMP` row to the connect-mode mapping table) is submitted separately to `debezium/debezium`.

### DIFF
--- a/src/main/java/io/debezium/connector/vitess/VitessValueConverter.java
+++ b/src/main/java/io/debezium/connector/vitess/VitessValueConverter.java
@@ -16,6 +16,7 @@ import java.time.format.DateTimeFormatter;
 import java.time.format.DateTimeFormatterBuilder;
 import java.time.format.DateTimeParseException;
 import java.time.temporal.ChronoField;
+import java.util.Date;
 import java.util.List;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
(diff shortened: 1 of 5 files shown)


jpechane (approved):
@mcrauwel LGTM, thanks!
Can be merged after the rebase.
Could you please just add and integration test to verify the timestamp with timezone mapping - the one from the other PR?

mcrauwel:
Rebased onto `main` (now includes #288) — conflicts in `VitessValueConverter` and the timestamp fixtures resolved so the `connect`-mode path (this PR) and the `ZonedTimestamp` ISO path (#288) coexist.

Also added a dedicated integration test `shouldReceiveChangesForInsertsWithTimestampWithTimeZoneMapping` verifying the timestamp-with-timezone (`ZonedTimestamp`) mapping end-to-end, as requested. Full timestamp IT matrix (default `adaptive_time_microseconds`, `connect`, `isostring`, zero-value variants) plus the converter unit tests pass against a live Vitess cluster.

Naros (approved):

Naros:
Thanks @mcrauwel, applied! :rocket:

PR #288, opened 2026-06-12
debezium/dbz#2063 Emit ISO 8601 strings for TIMESTAMP columns mapped to ZonedTimestamp

mcrauwel:
Fixes https://github.com/debezium/dbz/issues/2063

VStream emits MySQL `TIMESTAMP` values as raw UTC strings (`2020-02-13 01:02:03`), and `ZonedTimestamp.toIsoString` passes `String` input through verbatim — so the value carried under the `ZonedTimestamp` schema was never a valid ISO 8601 zoned timestamp, violating the type's contract (and differing from the binlog-based MySQL connector, which deserializes `TIMESTAMP` into a UTC `ZonedDateTime` and emits e.g. `2020-02-13T01:02:03Z`).

This change overrides `convertTimestampWithZone` in `VitessValueConverter` to parse the VStream string into a `ZonedDateTime` in UTC before delegating to the `JdbcValueConverters` implementation. The emitted value is now ISO 8601 in UTC, with fractional digits padded per the column's declared precision (`TIMESTAMP(3)` → `2020-02-13T01:02:03.123Z`), consistent with the MySQL connector and with the existing Vitess connector documentation (which already shows the ISO form).

**Compatibility notes:**
* This is a value-format change for all `time.precision.mode` settings that map `TIMESTAMP` to `ZonedTimestamp` (i.e. everything except `connect` after #287) — consumers that adapted to the raw format will see ISO 8601 strings instead. Likely deserves a release-note/breaking-change mention.
* The MySQL zero-date sentinel (`0000-00-00 00:00:00`), which previously passed through verbatim, now converts to null for optional columns and the epoch fallback for non-optional columns — mirroring the binlog connector's deserializer (null for values outside the legal `TIMESTAMP` range) and the existing zero-date `DATETIME` semantics in this connector.
* Verified with the full timestamp integration test matrix (default `adaptive_time_microseconds`, `connect`, `isostring`, all zero-value variants) against a live Vitess cluster; the existing IT expectations codified the raw pass-through values and are updated accordingly.

Note: this PR touches the same file as #287 (dbz#2060) but is independent; whichever merges second will need a trivial rebase — happy to do that promptly.

### DIFF
--- a/src/main/java/io/debezium/connector/vitess/VitessValueConverter.java
+++ b/src/main/java/io/debezium/connector/vitess/VitessValueConverter.java
@@ -9,7 +9,13 @@ import java.math.BigDecimal;
 import java.sql.Timestamp;
 import java.time.Duration;
 import java.time.LocalDate;
+import java.time.LocalDateTime;
 import java.time.ZoneOffset;
+import java.time.ZonedDateTime;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeFormatterBuilder;
+import java.time.format.DateTimeParseException;
+import java.time.temporal.ChronoField;
 import java.util.List;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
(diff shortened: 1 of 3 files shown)


jpechane:
@mcrauwel Applied, thanks!

PR #289, opened 2026-06-19
debezium/dbz#2126 Add protobuf and protoc overrides to upgrade

twthorn:
Bump version ahead of new vitess-grpc-client release (backward compatible still works with existing too)

Fixes https://github.com/debezium/dbz/issues/2126


jpechane:
@twthorn Applied, thanks.

PR #293, opened 2026-07-06
debezium/dbz#2191 Fix BIT column being silently dropped in Vitess connector

mcrauwel:
Fixes https://github.com/debezium/dbz/issues/2191

Two missing cases combine to silently drop any `BIT(N)` column from change events:

1. **`VitessType.resolve()`** — no `"BIT"` case in the type-name switch, so `BIT` columns fell through to `default` → JDBC type `OTHER` (1111).
2. **`ReplicationMessageColumnValueResolver.resolveValue()`** — no arm for `Types.BIT` (`-7`), so it fell through to `return null`, silently dropping the value.

**Fix:** add `"BIT" → Types.BIT` in `VitessType.resolve()`, and add `case Types.BIT: return value.asBytes()` alongside the existing `BLOB`/`BINARY` arm in `resolveValue()`. VStream already encodes `BIT` column values as raw bytes, and `JdbcValueConverters.convertBits` already maps `Types.BIT` to the correct Kafka Connect schema type (`BOOLEAN` for `BIT(1)`, `BYTES` for `BIT(N)`), so no further changes are needed.

Tests added in `VitessTypeTest` (resolves to `Types.BIT`) and `ReplicationMessageColumnValueResolverTest` (returns raw bytes).

### DIFF
--- a/src/main/java/io/debezium/connector/vitess/VitessValueConverter.java
+++ b/src/main/java/io/debezium/connector/vitess/VitessValueConverter.java
@@ -151,6 +151,27 @@ public class VitessValueConverter extends JdbcValueConverters {
         return matches(typeName, Query.Type.TIMESTAMP.name());
     }
 
+    @Override
+    protected ValueConverter convertBits(Column column, Field fieldDefn) {
+        if (column.length() <= 1) {
+            // BIT(1) - VStream sends raw bytes; convert to boolean as JdbcValueConverters expects
+            return (data) -> {
+                if (data instanceof byte[]) {
+                    byte[] bytes = (byte[]) data;
+                    return bytes.length > 0 && bytes[0] != 0;
+                }
+                return super.convertBits(column, fieldDefn).convert(data);
+            };
+        }
+        // BIT(N) - VStream sends raw bytes; return as-is for Kafka Connect BYTES schema
+        return (data) -> {
+            if (data instanceof byte[]) {
+                return data;
+            }
+            return super.convertBits(column, fieldDefn).convert(data);
+        };
+    }
+
     // Convert Java value to Kafka Connect value.
     @Override
     public ValueConverter converter(Column inputColumn, Field fieldDefn) {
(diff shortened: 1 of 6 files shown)


src/main/java/io/debezium/connector/vitess/VitessValueConverter.java
```diff
@@ -151,6 +151,27 @@ private static boolean isTimestamp(String typeName) {
...
+                return super.convertBits(column, fieldDefn).convert(data);
+            };
+        }
+        // BIT(N) - VStream sends raw bytes; return as-is for Kafka Connect BYTES schema
+        return (data) -> {
+            if (data instanceof byte[]) {
+                return data;
+            }
+            return super.convertBits(column, fieldDefn).convert(data);
+        };
```
Naros:
```suggestion
        return (data) -> {
            if (data instanceof byte[] bytes) {
              if (column.length() <= 1) {
                  return bytes.length > 0 && bytes[0] != 0;
              }
              return bytes;
            }
            return super.convertBits(column, fieldDefn).convert(data);
        };
```
What about condensing logic?

Naros (commented):
LGTM, I've just left one improvement request for the converter.

src/main/java/io/debezium/connector/vitess/VitessValueConverter.java
mcrauwel: Good call, done in 14e860d — applied with the project's 4-space indent so the formatter check stays happy. Thanks!

src/main/java/io/debezium/connector/vitess/VitessValueConverter.java
```diff
@@ -153,20 +153,14 @@ private static boolean isTimestamp(String typeName) {
...
     protected ValueConverter convertBits(Column column, Field fieldDefn) {
-        if (column.length() <= 1) {
-            // BIT(1) - VStream sends raw bytes; convert to boolean as JdbcValueConverters expects
-            return (data) -> {
-                if (data instanceof byte[]) {
-                    byte[] bytes = (byte[]) data;
+        // VStream sends raw bytes; convert BIT(1) to boolean as JdbcValueConverters expects,
+        // return BIT(N) as-is for Kafka Connect BYTES schema
+        return (data) -> {
+            if (data instanceof byte[] bytes) {
```
Naros, line 159: Even better, you added the `instanceof` pattern matching 🎉

Naros (approved):

Naros:

PR #295, opened 2026-08-21
debezium/dbz#2479 debezium/dbz#2480 Fix inherit.epoch validation and gRPC header value parsing

### LINKED ISSUE #2479: Vitess connector: inherit.epoch validation reports problems but never fails (validateInheritEpoch always returns 0)
**Version:** main (3.7.0-SNAPSHOT); present since `vitess.inherit.epoch` was introduced.

`VitessConnectorConfig.validateInheritEpoch` reports a validation problem via `problems.accept(...)` but unconditionally `return 0`:

https://github.com/debezium/debezium-connector-vitess/blob/610950b/src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java#L723-L730

Debezium `Field` validators return the number of problems found, so returning 0 tells the framework the config is valid. Result: `vitess.inherit.epoch=true` without `transaction.metadata.factory=VitessOrderedTransactionMetadataFactory` passes validation and the connector starts, instead of being rejected — the problem message is effectively dead code. Compare `validateTimePrecisionMode` in the same class, which correctly returns `1`.

Will submit a PR.

### LINKED ISSUE #2480: Vitess connector: vitess.grpc.headers silently drops headers whose value contains a colon

`VitessConnectorConfig.getGrpcHeaders` splits each header on `:` without a limit:

https://github.com/debezium/debezium-connector-vitess/blob/610950b/src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java#L773-L781

Any header whose **value** contains a colon (e.g. `authorization:Bearer x:y`, URLs, timestamps) splits into more than 2 parts, fails the `keyAndValue.length == 2` check, and the whole header is dropped with only a WARN log. HTTP/gRPC header values may legitimately contain colons — only the first colon separates the key.

Fix: `header.split(":", 2)`, so the value keeps any further colons.

Will submit a PR.

mcrauwel:
Fixes https://github.com/debezium/dbz/issues/2479
Fixes https://github.com/debezium/dbz/issues/2480

Two small `VitessConnectorConfig` bugs:

1. **`validateInheritEpoch`** reported its problem via `problems.accept(...)` but always returned `0`. Field validators return the number of problems found, so the configuration was treated as valid and the connector started with `vitess.inherit.epoch=true` but without `VitessOrderedTransactionMetadataFactory` — the validation message was effectively dead code. Now returns the problem count (matching `validateTimePrecisionMode`), and the factory comparison is null-safe.

2. **`getGrpcHeaders`** split each configured header on every colon, so any header whose *value* contains a colon (bearer tokens, URLs, timestamps) failed the `length == 2` check and was silently dropped with only a WARN log. Now splits on the first colon only (`split(":", 2)`); entries without any colon are still rejected with the warning.

Unit tests added for both: validation fail/pass pairs for `inherit.epoch`, header values with embedded colons, and the no-colon rejection path.

**AI disclosure** (per `AI_USAGE_POLICY.md`): I used Claude Code to help draft this fix and its tests. I reviewed the change and ran the test suite locally before submitting.

### DIFF
--- a/src/test/java/io/debezium/connector/vitess/VitessConnectorConfigTest.java
+++ b/src/test/java/io/debezium/connector/vitess/VitessConnectorConfigTest.java
@@ -305,4 +305,27 @@ public class VitessConnectorConfigTest {
         assertThat(problems).isEmpty();
     }
 
+    @Test
+    public void shouldGetGrpcHeadersWithColonsInValue() {
+        Configuration configuration = TestHelper.defaultConfig()
+                .with(VitessConnectorConfig.GRPC_HEADERS, "authorization:Bearer a:b:c,x-custom-header:value")
+                .build();
+        VitessConnectorConfig connectorConfig = new VitessConnectorConfig(configuration);
+        assertThat(connectorConfig.getGrpcHeaders())
+                .containsEntry("authorization", "Bearer a:b:c")
+                .containsEntry("x-custom-header", "value")
+                .hasSize(2);
+    }
+
+    @Test
+    public void shouldSkipGrpcHeadersWithoutColon() {
+        Configuration configuration = TestHelper.defaultConfig()
+                .with(VitessConnectorConfig.GRPC_HEADERS, "not-a-header,x-custom-header:value")
+                .build();
+        VitessConnectorConfig connectorConfig = new VitessConnectorConfig(configuration);
+        assertThat(connectorConfig.getGrpcHeaders())
+                .containsEntry("x-custom-header", "value")
+                .hasSize(1);
+    }
+
 }
--- a/src/test/java/io/debezium/connector/vitess/VitessConnectorConfigTest.java
+++ b/src/test/java/io/debezium/connector/vitess/VitessConnectorConfigTest.java
@@ -19,6 +19,7 @@ import org.junit.jupiter.api.Test;
 import io.debezium.config.CommonConnectorConfig;
 import io.debezium.config.Configuration;
 import io.debezium.connector.vitess.pipeline.txmetadata.VitessOrderedTransactionMetadataFactory;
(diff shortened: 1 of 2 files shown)


src/test/java/io/debezium/connector/vitess/VitessConnectorConfigTest.java
```diff
@@ -282,4 +282,50 @@ public void shouldGetConnectorGenerationDefaultValue() {
         assertThat(connectorConfig.getConnectorGeneration()).isEqualTo(0);
     }
 
+    @Test
+    public void shouldFailInheritEpochValidationWithoutOrderedTransactionMetadataFactory() {
```
Naros, line 288: For these new tests, could you add the `@FixFor("debezium/dbz#2479")` annotation?

Naros (commented):
Hi @mcrauwel the changes LGTM, just one small comment.

src/test/java/io/debezium/connector/vitess/VitessConnectorConfigTest.java
mcrauwel: Done in 753bd2f — added `@FixFor` to all four new tests (2479 on the validation pair, 2480 on the header pair). Thanks for the review!

mcrauwel:
@Naros btw, you may expect a bunch of PR from me in the coming days/weeks, we are working our way through bugfixes done on our fork of the connector to bring our fork as close as possible to the upstream connector.
See our [upstreaming plan](https://github.com/planetscale/debezium-connector-planetscale/blob/mcrauwel/dev-upstreaming-plan/.dev/upstreaming-plan.md)

Naros (approved):

Naros:

PR #298, opened 2026-09-02
debezium/dbz#2545 debezium/dbz#2546 Fix gRPC channel lifecycle and vgtid reference-equality bugs

### LINKED ISSUE #2545: Vitess connector: gRPC channels leaked on repeated connect, and close() throws NPE when never connected

Two related lifecycle bugs in `VitessReplicationConnection`:

1. **Channel leak** — `execute(String)`, `execute(String, String)` and `startStreaming(...)` each create a new `ManagedChannel` but store it with `managedChannel.compareAndSet(null, channel)`, which only retains the **first** channel ever created:
https://github.com/debezium/debezium-connector-vitess/blob/9e5232d/src/main/java/io/debezium/connector/vitess/connection/VitessReplicationConnection.java#L73-L74
Every subsequent call (task restarts, re-streaming after errors, metadata queries) creates a channel that is used for the RPC but never referenced again — `close()` only shuts down the first one. Each leaked `ManagedChannel` holds threads and sockets, and gRPC logs `ManagedChannel allocation site ... not shut down properly` garbage-collection warnings.

2. **NPE in `close()`** — `managedChannel.get().shutdownNow()` throws `NullPointerException` when no channel was ever created (startup failure before connecting):
https://github.com/debezium/debezium-connector-vitess/blob/9e5232d/src/main/java/io/debezium/connector/vitess/connection/VitessReplicationConnection.java#L391
During error handling/shutdown this NPE can mask the original failure.

Fix: swap with `getAndSet` and shut down the previous channel; null-guard `close()`. Will submit a PR.

### LINKED ISSUE #2546: Vitess connector: vgtid configuration compared with reference equality (==) instead of equals

`VitessReplicationConnection.defaultVgtid` compares the configured vgtid string against the `Vgtid.CURRENT_GTID` / `Vgtid.EMPTY_GTID` constants with `==`:
https://github.com/debezium/debezium-connector-vitess/blob/9e5232d/src/main/java/io/debezium/connector/vitess/connection/VitessReplicationConnection.java#L449
https://github.com/debezium/debezium-connector-vitess/blob/9e5232d/src/main/java/io/debezium/connector/vitess/connection/VitessReplicationConnection.java#L466-L467

This works today only by accident: `VitessConnectorConfig.getVgtid()` happens to return those exact constant references for the default / `snapshot.mode=initial` paths. A value that reaches this code as an equal-but-distinct string breaks it — concretely, an explicit `vitess.vgtid=""`:

- **with `vitess.shard` set** — both `==` checks fail, the code falls through to `Vgtid.of("")` and the task crashes on the JSON parse;
- **without `vitess.shard`** — the `== EMPTY_GTID` check fails and the connector **silently** builds a `current`-position vgtid instead of the requested from-the-beginning read, skipping the copy phase entirely.

Fix: compare with `equals` (constant-first). Will submit a PR.

mcrauwel:
Fixes https://github.com/debezium/dbz/issues/2545
Fixes https://github.com/debezium/dbz/issues/2546

Two connection-level bugs in `VitessReplicationConnection`:

1. **gRPC channel lifecycle** — `execute(...)` and `startStreaming(...)` each created a new `ManagedChannel` but stored it with `compareAndSet(null, channel)`, which retains only the first channel ever created: every later connect (task restart, re-streaming after an error, metadata query) leaked a channel with its threads and sockets, and `close()` only shut down the first one. The three call sites now go through a `replaceManagedChannel()` helper that swaps with `getAndSet` and shuts down the previous channel. `close()` also threw `NullPointerException` when no channel was ever created (startup failure before connecting), masking the original error during shutdown — it now null-guards and uses a single snapshot of the reference for both `shutdownNow()` and `awaitTermination(...)`.

2. **vgtid reference equality** — `defaultVgtid` compared the configured vgtid string against `Vgtid.CURRENT_GTID` / `Vgtid.EMPTY_GTID` with `==`, which only works because `getVgtid()` happens to return those exact constant references on the default paths. An explicit `vitess.vgtid=""` either crashed in `Vgtid.of("")` (with `vitess.shard` set) or **silently** built a `current`-position vgtid instead of the requested from-the-beginning read (without `vitess.shard`). Now compares with constant-first `equals`.

Unit tests added (`@FixFor` annotated): `close()` on a never-connected connection no longer throws, and an explicit empty vgtid with shards builds the correct from-the-beginning vgtid — the empty string is constructed at runtime so it is an equal-but-distinct instance, like a value parsed from user configuration. Both tests fail on current `main` and pass with the fixes.

**AI disclosure** (per `AI_USAGE_POLICY.md`): I used Claude Code to help draft this fix and its tests. I reviewed the change and ran the test suite locally before submitting.

### DIFF
--- a/src/main/java/io/debezium/connector/vitess/connection/VitessReplicationConnection.java
+++ b/src/main/java/io/debezium/connector/vitess/connection/VitessReplicationConnection.java
@@ -461,7 +461,7 @@ public class VitessReplicationConnection implements ReplicationConnection {
             // If offset storage per task is disabled, then find the vgtid elsewhere
             if (config.getShard() == null || config.getShard().isEmpty()) {
                 // This case is not supported by the Vitess, so our workaround is to get all the shards from vtgate.
-                if (config.getVgtid() == Vgtid.EMPTY_GTID) {
+                if (Vgtid.EMPTY_GTID.equals(config.getVgtid())) {
                     List<String> shards = new VitessMetadata(config).getShards();
                     List<String> gtids = Collections.nCopies(shards.size(), config.getVgtid());
                     vgtid = buildVgtid(config.getKeyspace(), shards, gtids);
@@ -478,8 +478,8 @@ public class VitessReplicationConnection implements ReplicationConnection {
                 List<String> shards = config.getShard();
                 String vgtidString = config.getVgtid();
                 List<String> gtids;
-                if (vgtidString == Vgtid.CURRENT_GTID ||
-                        vgtidString == Vgtid.EMPTY_GTID) {
+                if (Vgtid.CURRENT_GTID.equals(vgtidString) ||
+                        Vgtid.EMPTY_GTID.equals(vgtidString)) {
                     gtids = Collections.nCopies(shards.size(), vgtidString);
                     vgtid = buildVgtid(config.getKeyspace(), shards, gtids);
                 }
--- a/src/main/java/io/debezium/connector/vitess/connection/VitessReplicationConnection.java
+++ b/src/main/java/io/debezium/connector/vitess/connection/VitessReplicationConnection.java
@@ -70,8 +70,7 @@ public class VitessReplicationConnection implements ReplicationConnection {
      */
     public Vtgate.ExecuteResponse execute(String sqlStatement) {
         LOGGER.debug("Executing sqlStament {}", sqlStatement);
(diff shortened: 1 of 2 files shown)


Naros (approved):

Naros:
Thanks @mcrauwel, great work; applied! :tada:

PR #299, opened 2026-09-02
debezium/dbz#2548 Convert zero-day DATETIME values to null instead of crashing the task

### LINKED ISSUE #2548: Vitess connector: zero-day DATETIME values (e.g. 2024-01-00) crash the task; only zero-month is handled

`VitessValueConverter.stringToTimestamp` nulls out invalid zero-datetimes with the regex `^\\d{4}-00-00.*$`, which only matches a zero **month**:

https://github.com/debezium/debezium-connector-vitess/blob/9e5232d/src/main/java/io/debezium/connector/vitess/VitessValueConverter.java#L437-L443

MySQL (with the default `NO_ZERO_IN_DATE` not enforced) also permits a valid month with a zero **day**, e.g. `2024-01-00 10:00:00`. Such a value falls through to `Timestamp.valueOf(...)`, which throws `IllegalArgumentException` and kills the task.

The companion `stringToLocalDate` already handles this correctly — it extracts year/month/day with `DATE_FIELD_PATTERN` and treats `month == 0 || day == 0` as the invalid-value case. `stringToTimestamp` should do the same.

Will submit a PR.

mcrauwel:
Fixes https://github.com/debezium/dbz/issues/2548

`stringToTimestamp` nulls out invalid zero-datetimes with the regex `^\d{4}-00-00.*$`, which only matches a zero **month**. MySQL also permits a valid month with a zero **day** (e.g. `2024-01-00 10:00:00`); such a value fell through to `Timestamp.valueOf(...)`, which throws `IllegalArgumentException` and kills the task.

The fix mirrors the companion `stringToLocalDate`: extract year/month/day with the existing `DATE_FIELD_PATTERN` and treat `month == 0 || day == 0` as the invalid-value case (WARN + null). Strings that don't match the pattern fall through to `Timestamp.valueOf` unchanged.

Unit test added (`@FixFor` annotated) covering `0000-00-00`, zero-month, and zero-day datetimes — it fails on current `main` with the `IllegalArgumentException` above and passes with the fix.

**AI disclosure** (per `AI_USAGE_POLICY.md`): I used Claude Code to help draft this fix and its test. I reviewed the change and ran the test suite locally before submitting.

### DIFF
--- a/src/main/java/io/debezium/connector/vitess/VitessValueConverter.java
+++ b/src/main/java/io/debezium/connector/vitess/VitessValueConverter.java
@@ -435,9 +435,16 @@ public class VitessValueConverter extends JdbcValueConverters {
      * @return The java.sql.Timestamp
      */
     public static Timestamp stringToTimestamp(String datetimeString) {
-        if (datetimeString.matches("^\\d{4}-00-00.*$")) {
-            INVALID_VALUE_LOGGER.warn("Invalid value '{}' stored in column converted to null value", datetimeString);
-            return null;
+        final Matcher matcher = DATE_FIELD_PATTERN.matcher(datetimeString);
+        if (matcher.find()) {
+            final int month = Integer.parseInt(matcher.group(2));
+            final int day = Integer.parseInt(matcher.group(3));
+            if (month == 0 || day == 0) {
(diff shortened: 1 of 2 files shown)


Naros (approved):

Naros:
Thanks @mcrauwel, thanks for the PR - applied :tada:

PR #300, opened 2026-09-02
debezium/dbz#2547 Add vitess.cells option to pin the VStream to specific cells

### LINKED ISSUE #2547: Vitess connector: expose vitess.cells to pin the VStream to specific cell(s)
**Feature request** for debezium-connector-vitess.

vtgate's `VStream` API accepts `VStreamFlags.cells` — a comma-separated list of cells (or cell aliases) that restricts which tablets serve the stream. The connector builds `VStreamFlags` in `VitessReplicationConnection.startStreaming` but does not expose this flag:

https://github.com/debezium/debezium-connector-vitess/blob/9e5232d/src/main/java/io/debezium/connector/vitess/connection/VitessReplicationConnection.java#L302-L306

For multi-cell Vitess deployments this is the standard locality control: without it the VStream may be served from tablets in a remote cell, adding cross-cell traffic and latency. `vitess-grpc-client` already exposes `VStreamFlags.Builder#setCells`, so this is a small config-plumbing change: a new optional `vitess.cells` STRING option, registered in `CONFIG_DEFINITION`, and applied to the flags builder when non-empty.

Will submit a PR.

mcrauwel:
Fixes https://github.com/debezium/dbz/issues/2547

vtgate's `VStream` API accepts `VStreamFlags.cells` — a comma-separated list of cells (or cell aliases) that restricts which tablets serve the stream. This adds a new **optional** `vitess.cells` STRING option, registered in `CONFIG_DEFINITION` (CONNECTOR_ADVANCED group, next to the other VStream flag `vitess.stop_on_reshard`), and applies it to the flags builder only when non-empty — when unset, vtgate's default cell selection applies, so existing configurations are unaffected.

Motivation: in multi-cell Vitess deployments this is the standard locality control; without it the VStream may be served from tablets in a remote cell, adding cross-cell traffic and latency.

Unit tests added (`@FixFor` annotated): option round-trip, null default, and exposure via `CONFIG_DEFINITION`/`ALL_FIELDS`.

Happy to follow up with a docs PR to `debezium/debezium` for the connector property table — let me know if that should land together with this or after.

**AI disclosure** (per `AI_USAGE_POLICY.md`): I used Claude Code to help draft this change and its tests. I reviewed the change and ran the test suite locally before submitting.

### DIFF
--- a/src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
+++ b/src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
@@ -326,7 +326,20 @@ public class VitessConnectorConfig extends RelationalDatabaseConnectorConfig {
             .withWidth(Width.MEDIUM)
             .withImportance(ConfigDef.Importance.MEDIUM)
             .withDescription("Comma-separated list of Vitess cells (or cell aliases) the VStream"
-                    + " should be served from. If not set, vtgate applies its default cell selection.");
+                    + " should be served from. If not set, vtgate defaults to its own local cell."
+                    + " Only applies when '" + VITESS_CONFIG_GROUP_PREFIX + "tablet.type' is REPLICA or RDONLY;"
+                    + " for MASTER the shard primary is used regardless of cell.");
+
+    public static final Field CELL_PREFERENCE = Field.create(VITESS_CONFIG_GROUP_PREFIX + "cell.preference")
+            .withDisplayName("VStream cell preference")
+            .withType(Type.STRING)
+            .withWidth(Width.MEDIUM)
+            .withImportance(ConfigDef.Importance.MEDIUM)
+            .withValidation(VitessConnectorConfig::validateCellPreference)
+            .withDescription("Controls how vtgate's tablet picker treats the cells listed in '" + VITESS_CONFIG_GROUP_PREFIX + "cells':"
+                    + " 'preferlocalwithalias' (vtgate's default) prefers tablets in vtgate's local cell and its alias before the other listed cells,"
+                    + " while 'onlyspecified' restricts tablet selection to exactly the listed cells (no local-cell fallback),"
+                    + " which is required to pin the VStream to specific cells.");
 
     public static final Field INHERIT_EPOCH = Field.create(VITESS_CONFIG_GROUP_PREFIX + "inherit.epoch")
             .withDisplayName("Inherit epoch")
@@ -563,6 +576,7 @@ public class VitessConnectorConfig extends RelationalDatabaseConnectorConfig {
             .group(Field.Group.CONNECTOR_ADVANCED,
                     STOP_ON_RESHARD_FLAG,
                     CELLS,
+                    CELL_PREFERENCE,
                     INHERIT_EPOCH,
                     SHARD_EPOCH_MAP,
                     KEEPALIVE_INTERVAL_MS,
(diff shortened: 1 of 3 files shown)


src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
```diff
@@ -320,6 +320,14 @@ public static BigIntUnsignedHandlingMode parse(String value, String defaultValue
             .withDescription("Control StopOnReshard VStream flag."
                     + " If set true, the old VStream will be stopped after a reshard operation.");
 
+    public static final Field CELLS = Field.create(VITESS_CONFIG_GROUP_PREFIX + "cells")
```
Naros, line 387: Hi @mcrauwel, do you plan to introduce cell preferences (preferred vs. exclusive)?

mcrauwel: Hey @Naros good call out, without it, `vitess.cells` can only widen the fallback pool: vtgate's tablet picker defaults to `preferlocalwithalias`, which always prefers vtgate's local cell (and its alias) before the listed cells. Exposing `cell_preference` with `onlyspecified` is what makes it possible to actually pin the VStream to specific cells.

mcrauwel: Pushed a follow-up commit (0b1d741) that adds `vitess.cell.preference`, passed through to `VStreamFlags.cell_preference`. The value is validated against the two options vtgate accepts — `preferlocalwithalias` and `onlyspecified` — case-insensitively, matching vtgate's own parsing. I also documented on both options that they only take effect when `vitess.tablet.type` is REPLICA or RDONLY, since for MASTER vtgate always picks the shard primary regardless of cell. The docs follow-up to debezium/debezium mentioned in the PR description would then cover both properties.

src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
```diff
@@ -320,6 +320,27 @@ public static BigIntUnsignedHandlingMode parse(String value, String defaultValue
...
+            .withWidth(Width.MEDIUM)
+            .withImportance(ConfigDef.Importance.MEDIUM)
+            .withDescription("Comma-separated list of Vitess cells (or cell aliases) the VStream"
+                    + " should be served from. If not set, vtgate defaults to its own local cell."
+                    + " Only applies when '" + VITESS_CONFIG_GROUP_PREFIX + "tablet.type' is REPLICA or RDONLY;"
+                    + " for MASTER the shard primary is used regardless of cell.");
+
+    public static final Field CELL_PREFERENCE = Field.create(VITESS_CONFIG_GROUP_PREFIX + "cell.preference")
+            .withDisplayName("VStream cell preference")
+            .withType(Type.STRING)
```
jpechane: Can we use `withEnum` and define and handle the option es java enum?

jpechane (changes requested):
@mcrauwel Overall, LGTM, thanks. But I beleive the preference should be handled as an enum.

src/main/java/io/debezium/connector/vitess/VitessConnectorConfig.java
mcrauwel: Hey @jpechane, good point! Done in 899391e — added a `CellPreference` enum (same pattern as `BigIntUnsignedHandlingMode`) and wired it up with `withEnum`, which also let me drop the hand-rolled validator. It defaults to `PREFER_LOCAL_WITH_ALIAS`, which is what vtgate does anyway when the flag isn't set, so behavior only changes when you actually configure it.

mcrauwel:
> @mcrauwel Overall, LGTM, thanks. But I beleive the preference should be handled as an enum.

Resolved this in the thread. Moved the handling to be an enum.

Naros (approved):

mcrauwel:
@jpechane I fixed your requested changes last week, can you have another look :)

mcrauwel:
@Naros @jpechane just a follow up ping here :) thanks

Naros:

